PR #455, opened 2020-12-09
Fixed missing droplet property in example

scotchneat:
While working on the contract tests, I found an issue with the GET Droplet response example


scotchneat:
I'm not sure why the diff is so all over the place but all I did was move all the example properties under a `droplet` property.

ChiefMateStarbuck:
LGTM

ChiefMateStarbuck (approved):

PR #459, opened 2020-12-14
Update floating IP responses.

andrewsomething:
This includes the additional `links` object in the POST `/v2/floating_ips` response missing from the public documentation as well as a fix in the POST `/v2/floating_ips/{ip}/actions` request to pass Prism validation.


bentranter (dismissed):
LGTM (pending CI)

specification/resources/floating_ips/models/floating_ip_actions.yml
```diff
@@ -1,21 +1,31 @@
...
   properties:
     type:
-      type: string
-      enum:
-      - assign
-      - unassign
-      example: assign
-      description: The type of action to initiate for the floating IP.
+      $ref: '#/floating_ip_action_type'
+  additionalProperties: false
```
andrewsomething, line 14: This is something we might need to use in other places where we make use of discriminators. Without forbidding additional properties `{"type": "unassign"}` and `{"type": "assign","droplet_id":123456}` both match the unassign schema.

    [VALIDATOR] ✖  error     Violation: request.body should match exactly one schema in oneOf

bentranter: That's interesting, I hadn't seen `additionalProperties` before. That makes sense, and yeah I think this will show up in other places as we test this.

bentranter (approved):

PR #467, opened 2021-01-14
apps: Remove unused pagination responses.

andrewsomething:
These models are not actually used. The standard DO API pagination models are ref-ed.

```
$ grep -r apps_pagination_response.yml specification/
$ grep -r apps_pagination_request.yml specification/
```


ChiefMateStarbuck (approved):
LGTM

PR #468, opened 2021-01-14
apps: Remove unused health models.

andrewsomething:
This operation and corresponding models are not routed publicly.


ChiefMateStarbuck (approved):
LGTM

PR #470, opened 2021-01-14
apps: Remove unused git models.

andrewsomething:
Removes additional models are not used in any public routes.


ChiefMateStarbuck (approved):
LGTM

PR #473, opened 2021-01-25
domain record: Fix table in tag description.

andrewsomething:
The table in the domain records tag description seems to have combined both of the tabled from:

https://developers.digitalocean.com/documentation/v2/#domain-records

image

The third and forth columns are supposed to specify which attributes are required for each record type, but it is missing the column with attributes. I'm not entire sure what was intended here, but since the discriminator now shows this info, I think we can just remove it.


ChiefMateStarbuck (approved):
👍🏾

PR #475, opened 2021-02-03
boolean is not a valid format.

andrewsomething:
When experimenting with some code generators, I've notice that some warn or error on the usage of `format: boolean`. 

`format` extends `type` which is already `boolean` in these cases.

https://swagger.io/specification/#data-types


bentranter (approved):
LGTM! (Not sure what caused the upload failure, but doesn't seem like a blocker)

PR #479, opened 2021-02-08
Add new Kubernetes associated resources endpoints

andrewsomething:
https://developers.digitalocean.com/documentation/changelog/api-v2/doks-cascading-deletes/

### DIFF
--- a/specification/resources/kubernetes/models/associated_kubernetes_resources.yml
+++ b/specification/resources/kubernetes/models/associated_kubernetes_resources.yml
@@ -1,0 +1,30 @@
+type: object
+description: An objects containing the IDs of resources associated with a
+  Kubernetes cluster.
+properties:
+  load_balancers:
+    type: array
+    items:
+      type: string
+    example:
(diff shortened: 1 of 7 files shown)


specification/resources/kubernetes/models/associated_kubernetes_resources.yml
```diff
@@ -0,0 +1,30 @@
+type: object
+description: An objects containing the IDs of resources associated with a
```
bentranter, line 2: Should this be,
```suggestion
description: An object containing the IDs of resources associated with a
```

bentranter (commented):
One potential typo, other than that LGTM!

specification/resources/kubernetes/models/associated_kubernetes_resources.yml
andrewsomething: Good catch!

bentranter (approved):
LGTM!

PR #480, opened 2021-02-11
apps request/response clean up

ChiefMateStarbuck:


specification/resources/apps/models/app_component_base.yml
```diff
@@ -9,49 +9,14 @@ properties:
     description: The name. Must be unique across all components within the same app.
     example: api
 
-  git:
-    $ref: apps_git_source_spec.yml
-
   github:
     $ref: apps_github_source_spec.yml
 
-  gitlab:
```
andrewsomething, line 18: It's not clear to me why these are being removed here as well the fields in other files.  The are valid parts of the App Specification.

https://www.digitalocean.com/docs/app-platform/references/app-specification-reference/
https://pkg.go.dev/github.com/digitalocean/godo#AppSpec

I think the models are fine as is. I think the create app operation could use a example for the request body. Without a specific example, the default examples for each possible field get used. 

https://github.com/digitalocean/openapi/blob/main/specification/resources/apps/create_app.yml#L15

andrewsomething (changes requested):

scotchneat (approved):
Looks good

docker-compose.yaml
```diff
@@ -16,4 +16,4 @@ services:
     volumes:
       - ./specification:/prism
     ports:
-      - "8080:4010"
\ No newline at end of file
+      - "8090:4010"
```
andrewsomething, line 19: Why the change here?

andrewsomething (approved):
Looks great!

I did have one question in-line, but not necessarily blocking. Just want to understand it.

docker-compose.yaml
ChiefMateStarbuck: I found that running both a spec preview and a prod mock at the same time can help with visibility into any code changes. since both the preview and the prod mock both ran on port 8080, changing this port helps allows for both to be run simultaneously.

andrewsomething: :+1:

PR #483, opened 2021-02-18
Resolving errors pertaining to "floating_ip" clientgen

danaelhe:
Was running openapi generator and ran into the following exception:
Exception: 'null' defines discriminator 'type', but the referenced schema 'floating_ip_action_unassign' is incorrect. invalid type for type, set it to string

Solution: must specify the discriminator mapping in the base schema, which in this case is "floating_ip_action_type"

Was able to move past that exception and on to another, which will be resolved in a different pr.

### DIFF
--- a/specification/resources/floating_ips/models/floating_ip_actions.yml
+++ b/specification/resources/floating_ips/models/floating_ip_actions.yml
@@ -1,31 +1,34 @@
 floating_ip_action_type:
-  type: string
-  enum:
-  - assign
-  - unassign
-  example: assign
-  description: The type of action to initiate for the floating IP.
+  type: object
+  required:
+    - type
+  discriminator:
+    propertyName: type
+    mapping:
+      ASSIGN: '#/floating_ip_action_unassign'
+      UNASSIGN: '#/floating_ip_action_assign'
 
 floating_ip_action_unassign:
-  type: object
-  properties:
-    type:
-      $ref: '#/floating_ip_action_type'
-  additionalProperties: false
-  required:
-  - type
+  allOf:
+    - $ref: '#/floating_ip_action_type'
+    - type: object
+      required:
+        - type
+      properties:
+        type:
+          type: string
 
 floating_ip_action_assign:
-  type: object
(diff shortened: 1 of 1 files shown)


ChiefMateStarbuck (dismissed):
🥇

specification/resources/floating_ips/models/floating_ip_actions.yml
```diff
@@ -1,31 +1,34 @@
 floating_ip_action_type:
-  type: string
-  enum:
-  - assign
-  - unassign
-  example: assign
-  description: The type of action to initiate for the floating IP.
```
andrewsomething, line 7: We're loosing the description for the `type` attribute here. Can we still include it?

specification/resources/floating_ips/models/floating_ip_actions.yml
```diff
@@ -1,31 +1,34 @@
...
   required:
-  - type
-
-floating_ip_action_assign:
-  type: object
-  properties:
-    type:
-      $ref: '#/floating_ip_action_type'
+    - type
+  discriminator:
```
andrewsomething, line 5:
```suggestion
  properties:
    type:
      type: string
      description: The type of action to initiate for the floating IP.
  discriminator:
```

Since we're using allOf with `$ref: '#/floating_ip_action_type'` in the other objects, if we add this here we do not need to repeat 

```
        type:
          type: string
```

in both of them. It also retains the description for `type`.

andrewsomething (commented):
Looking good! Confirms that it resolves the openapi-generator issue. :boom: 

I did have a suggestion inline.

danaelhe:
> Looking good! Confirms that it resolves the openapi-generator issue. 💥
> 
> I did have a suggestion inline.

ah, good catch. I've pushed another commit to resolve the duplication and maintain the description.

danaelhe:
> 🥇

thanks sam!!

andrewsomething (approved):
:shipit:  Looks good!

PR #485, opened 2021-02-19
Resolving errors with openapi-generator 5.0.0

danaelhe:
When inputting the DO openapi spec into the latest release of openapi-generator, I would get an exception with new_volume_ext4 and new_volume_xfs (a "filesystem_type" needs to be required and a type of string")

Solution: include discriminator in the base schema (volume_write_file_system_type).

This resolved the exceptions.

With this PR, the DO spec is accepted into the latest version of openapi-generator (5.0.0) exception/error free and is able to generate a python client.

### DIFF
--- a/specification/resources/volumes/models/attributes.yml
+++ b/specification/resources/volumes/models/attributes.yml
@@ -1,18 +1,26 @@
 volume_write_file_system_type:
-  type: string
-  description: >-
-    The name of the filesystem type to be used on the volume.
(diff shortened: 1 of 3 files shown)


danaelhe:
@andrewsomething created a new pr to resolve merge issues- thanks for pointing that out!

scotchneat (approved):
LGTM 👍🏽

PR #490, opened 2021-03-08
hiding app detect

ChiefMateStarbuck:
The apps team has requested that this endpoint be hidden in public documentation.

### DIFF
--- a/specification/DigitalOcean-public.v2.yaml
+++ b/specification/DigitalOcean-public.v2.yaml
@@ -458,9 +458,10 @@ paths:
     post:
       $ref: 'resources/apps/create_app.yml'
 
-  /v2/apps/detect:
-    post:
-      $ref: 'resources/apps/detect.yml'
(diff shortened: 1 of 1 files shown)


specification/DigitalOcean-public.v2.yaml
```diff
@@ -458,9 +458,10 @@ paths:
     post:
       $ref: 'resources/apps/create_app.yml'
 
-  /v2/apps/detect:
-    post:
-      $ref: 'resources/apps/detect.yml'
+        # Curently hiding this enpoint per request of apps team
+        #/v2/apps/detect:
+        #post:
+        #$ref: 'resources/apps/detect.yml'
```
scotchneat, line 464: Instead of commenting this, can we just remove it from the spec?

andrewsomething: Think we can just close this out now?

https://github.com/digitalocean/openapi/pull/500

PR #491, opened 2021-03-16
Example Code Migration

dbrian57:
What it says on the tin!


andrewsomething:
!deploy

specification/resources/apps/examples/curl/DUPE.yml
```diff
@@ -0,0 +1,2 @@
+lang: cURL
```
andrewsomething, line 1: There are a few of these `DUPE.yml` files that don't seem to be referenced anywhere. Something left over from some scripting?

specification/resources/droplets/examples/curl/post_droplet_action.yml
```diff
@@ -0,0 +1,3 @@
+lang: cURL
+source: |-
+  curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer b7d03a6947b217efb6f3ec3bd3504582" -d '{"type":"enable_backups"}' "https://api.digitalocean.com/v2/droplets/3164450/actions"
```
andrewsomething, line 3: One of the big differences between the OpenAPI spec and the current documentation is how these "action" endpoints are handled.

In the current docs, each "action" is its own entry even though they are all  POSTs to `/v2/droplets/<id>/actions`, just with different bodies. E.g:

https://developers.digitalocean.com/documentation/v2/#enable-backups
https://developers.digitalocean.com/documentation/v2/#disable-backups

and so on... OpenAPI handles this in a different way. For example, look at how you can toggle between different "types" and how that changes the other attributes.

https://api-engineering.nyc3.digitaloceanspaces.com/spec-ci/redoc-index.html?pr=491#operation/post_droplet_action

So we're losing some examples here. Might make sense to use the pattern you shared showing multiple examples for the various actions endpoints.

specification/resources/droplets/examples/go/post_droplet_action_by_tag.yml
```diff
@@ -0,0 +1,15 @@
...
+      "github.com/digitalocean/godo"
+  )
+
+  func main() {
+      pat := "mytoken"
+
+      client := godo.NewFromToken(pat)
+      ctx := context.TODO()
+
+      // ...
```
andrewsomething, line 14:
```suggestion
      action, _, err := client.DropletActions.PowerOffByTag(ctx, "awesome") 
```

andrewsomething (changes requested):
:tada: This is great stuff! Thanks for taking this on @dbrian57

I left a few comments inline, but this looks great.

specification/resources/1-clicks/examples/curl/1_click.yml
```diff
@@ -0,0 +1,3 @@
+lang: cURL
+source: |-
+  curl -X GET -H "Content-Type: application/json"  -H "Authorization: Bearer b7d03a6947b217efb6f3ec3bd3504582" "https://api.digitalocean.com/v2/1-clicks"
```
scotchneat, line 3: This is non-blocking and could be done in a subsequent change.
It might be helpful to split the curl commands into multiple lines.

```suggestion
  curl -X GET \
  -H "Content-Type: application/json"  \
  -H "Authorization: Bearer b7d03a6947b217efb6f3ec3bd3504582" \
  "https://api.digitalocean.com/v2/1-clicks" 
```

andrewsomething: I had the same thought. Think we can probably come up with some scripting to do this.

andrewsomething: Took a stab at it. This seems to do what we want:

```
grep -rl '\-H "Content-Type:' specification/*/*/examples/curl | xargs sed -i 's/-H "Content/\\\n    -H "Content/g'
grep -rl '\-H "Authorization:' specification/*/*/examples/curl | xargs sed -i 's/-H "Authorization/\\\n    -H "Authorization/g'
grep -rl "\-d '{" specification/*/*/examples/curl | xargs sed -i 's/-d/\\\n    -d/g'
grep -rl '"https://api' specification/*/*/examples/curl | xargs sed -i 's|"https://api|\\\n    "https://api|g'
```

specification/resources/droplets/examples/curl/post_droplet_action.yml
dbrian57: I've preserved the action examples in the way we discussed. I so wish there was a mapping function for `x-codeExample`s like there is for the `example` field.

scotchneat:
!deploy

andrewsomething (approved):
:shipit: Looks great!

PR #493, opened 2021-03-17
Use environment variables for tokens in Go and Ruby examples. 

andrewsomething:
Let's recommend good practices and make these consistent with the cURL examples.

### DIFF
--- a/specification/resources/certificates/examples/go/create_certificates.yml
+++ b/specification/resources/certificates/examples/go/create_certificates.yml
@@ -2,11 +2,13 @@ lang: Go
 source: |-
   import (
       "context"
+      "os"
+
       "github.com/digitalocean/godo"
   )
 
   func main() {
-      pat := "mytoken"
+      pat := os.Getenv("DIGITALOCEAN_TOKEN")
 
       client := godo.NewFromToken(pat)
       ctx := context.TODO()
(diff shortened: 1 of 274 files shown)


dbrian57 (approved):
lgtm! Thanks for cleaning this up.

scotchneat (approved):
This looks good. 
What do you think about changing `pat` to `token` in the go examples to be consistent across languages?

andrewsomething:
@scotchneat  Good idea. Updated.

specification/resources/certificates/examples/go/create_certificates.yml
```diff
@@ -2,26 +2,28 @@ lang: Go
...
   func main() {
-      pat := "mytoken"
+      token := os.Getenv("DIGITALOCEAN_TOKEN")
 
-      client := godo.NewFromToken(pat)
+      client := godo.NewFromToken(token)
       ctx := context.TODO()
 
-      key, err := ioutil.ReadFile("/path/to/privkey1.pem")
+      key, err := ioutil.ReadFile("/tokenh/to/privkey1.pem")
```
scotchneat, line 16: You could try matching `pat :=` and `(pat)` instead.

scotchneat (changes requested):
Oops. Looks like some extra lines were matched.

specification/resources/certificates/examples/go/create_certificates.yml
andrewsomething: :facepalm:

scotchneat (approved):
:shipit:

PR #494, opened 2021-03-18
tags: Fix regex for tag name in path.

andrewsomething:
The regex for tag name in path is missing `0` as a valid character. This also updates the regex used in the tag model so the two are consistent.

Fixing error found in contract testing:

```
=== RUN   TestTag_CreateShort/get_tag
    round_trippers.go:94: request_id=6c4bd201-ef6a-460b-818d-6ded21e0b8cb method=GET host=localhost:4010 path=/v2/tags/ct-107eb4dffbc62229 status_code=200 err=<nil> took=902.919554ms
    prism_proxy.go:208: [CONTRACT TEST VIOLATION] code=pattern message=should match pattern "^[a-zA-Z1-9_\-\:]+$" severity=Error location=request.path.tag_id
```

### DIFF
--- a/specification/resources/tags/models/tag.yml
+++ b/specification/resources/tags/models/tag.yml
@@ -1,31 +1,32 @@
 type: object
 
-description: >- 
-  A tag is a label that can be applied to a resource (currently Droplets, Images, Volumes, Volume Snapshots, and Database clusters) 
+description: >-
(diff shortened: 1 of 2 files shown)


scotchneat (approved):
👍🏽

Issue #498, opened 2021-04-01
No 400s or 403s, 404s or 429s in the definition


savage-alex:
Hi, Not an issue per say but some questions for the benefit of discussion:

What would you do if a consumer includes unexpected query params or a date time is in the wrong format? Have you thought of adding a 400 responses and outline how you could inform the consumer they made a mistake?

Is there a concept of AuthZ where the token may be legit but the client is not able to complete the opteration?

Also you have resources with path params but no 404 documented.

You are returning rate limit headers - this are brilliant but you are not including a 429 response to let people know - if u go over your SLA we are going to rate limit you the hard way :-)

Just some pointers. Really like the spectral rules.

andrewsomething:
Hey Alex!

Thanks for the great feedback. There is definitely a lot that can still be improved around error responses.

In this spec we are describing an API that has been around for sometime and is provided by a number of different micro-services. So the spec is not used to generate a server. Writing this spec is part of an effort to drive consistency between these services.

The actual API does return 429 responses when rate limits are reached and returns 400s/422s where appropriate. We'll want to add those here at some point to support more realistic mock servers, etc.

Though we should be including 404s already. Sounds like we need a Spectral rule to enforce that! :smile:

savage-alex:
adv-paths-with-parameters-with-404-response:
    description: All operations with path parameters MUST have a 404 response.
    message: >-
      {{path}} MUST have a 404 response. All operations with path parameters
      MUST have a 404 response.
    severity: error
    given: $.paths[?(@property.match(/.*\/{.*}.*/))].*
    then:
      field: responses
     
your function or truuthy here

savage-alex:
Do you also use 415 when people include response bodies that are in incorrect media types? (And spectral rule to make sure people do not add 415 for items that do not accept a response body)

savage-alex:
WOW you work fast!

PR #502, opened 2021-04-01
Add missing 404 responses and Spectral rule to enforce them.

andrewsomething:
The first commit adds a Spectral rule to enforce including 404 responses when a path requires a resource ID. The next commit fixes the violations reported by that rule.

Related to https://github.com/digitalocean/openapi/issues/498


bentranter (approved):
LGTM!

PR #507, opened 2021-04-13
Load Balancer Resize

dbrian57:
You can now resize load balancers via the API. I've updated the `size` field and examples to indicate that.


andrewsomething (approved):
:ship:  Thanks Dan!

PR #508, opened 2021-04-19
Ported Original API Intro Text to API Description

dbrian57:
Ported the [intro text from Developers](https://developers.digitalocean.com/documentation/v2/#introduction) into the spec's initial description field. Not a fan of the excessive whitespace, but I think we can fix that in the CSS on the docs site.

Thoughts welcome!


andrewsomething:
!deploy

specification/DigitalOcean-public.v2.yaml
```diff
@@ -3,9 +3,408 @@ openapi: 3.0.0
 info:
   title: DigitalOcean API
   version: '2.0'
-  description: >
+  description: |
```
andrewsomething, line 6: I wonder if this is something that can be ref-ed from another file?

```
  description:
    $ref: 'introduction.yaml'
```

scotchneat: Agreed. Rebilly takes that approach and it works well - https://github.com/Rebilly/api-definitions/blob/master/openapi/combined.yaml#L17.
They even take it further in their [despription.md](https://github.com/Rebilly/api-definitions/blob/master/openapi/description.md) and reference common errors nicely

andrewsomething: Nice! I didn't actually realize you could ref a markdown file directly. That would make it much easier to edit on its own.

dbrian57: That's a really good idea! I'll report back the results.

scotchneat:
Well, bummer. Sorry to lead you down that path @dbrian57 
It seems the Rebilly implementation just uses the Redoc tooling (openapi-cli) and don't leverage spectral like us. Redoc is running some custom spec parsing beyond what the actual OpenAPI v3 specification supports to enable this.

We might have to revisit using external markdowns some other time.

andrewsomething:
!deploy

andrewsomething:
!deploy

specification/description.yml
```diff
@@ -0,0 +1,401 @@
...
+
+  If you need access to the headers of a response through `curl`, you can pass
+  the `-i` flag to display the header information along with the body. If you
+  are only interested in the header, you can instead pass the `-I` flag, which
+  will exclude the response body entirely.
+
+  #### Set and Export your OAuth Token
+
+  ```
+  export TOKEN=your_token_here
```
andrewsomething, line 295:
```suggestion
  export DIGITALOCEAN_TOKEN=your_token_here
```

specification/description.yml
```diff
@@ -0,0 +1,401 @@
...
+  ### How to Authenticate with OAuth
+
+  In order to make an authenticated request, include a bearer-type
+  `Authorization` header containing your OAuth token. All requests must be
+  made over HTTPS.
+
+  #### Authenticate with a Bearer Authorization Header
+
+  ```
+  curl -X $HTTP_METHOD -H "Authorization: Bearer $TOKEN" "https://api.digitalocean.com/v2/$OBJECT"
```
andrewsomething, line 332:
```suggestion
  curl -X $HTTP_METHOD -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" "https://api.digitalocean.com/v2/$OBJECT"
```

specification/description.yml
```diff
@@ -0,0 +1,401 @@
...
+  export TOKEN=your_token_here
+  ```
+
+  #### Set and Export a Variable
+
+  ```
+  export DROPLET_ID=1111111
+  ```
+
+  ## OAuth Authentication
```
andrewsomething, line 304: This is just an idea and doesn't need to block this. Redoc creates a section for "Authentication" using the `securitySchemes`. That mostly empty section has felt a bit awkward to me.

https://api-engineering.nyc3.digitaloceanspaces.com/spec-ci/redoc-index.html?pr=508#section/Authentication

Wonder if we could rework the content here so that it shows up under the Redoc "Authentication" header.

https://github.com/digitalocean/openapi/blob/059a164acebf788fb555304a3b0f98fba3d2c05b/specification/DigitalOcean-public.v2.yaml#L1184-L1187

Looks like it can contain a `description` https://swagger.io/specification/#security-scheme-object

dbrian57: all for this and agree. Let me see what I can do

andrewsomething:
!deploy

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1185,5 +1185,105 @@ components:
...
+        tokens secure. In fact, upon generation, the web interface will only display
+        each token a single time in order to prevent the token from being compromised.
+
+        ### How to Authenticate with OAuth
+
+        In order to make an authenticated request, include a bearer-type
+        `Authorization` header containing your OAuth token. All requests must be
+        made over HTTPS.
+
+        #### Authenticate with a Bearer Authorization Header
```
andrewsomething, line 1214: Redoc's docs mention `<PullRight>` which would move this to the right hand example column, but it looks like this is only part of their hosted version: 

https://redoc.ly/docs/api-reference-docs/embedded-markdown/#pull-right
https://redoc.ly/redoc-vs-reference/

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1185,5 +1185,105 @@ components:
...
+
+        ```
+        curl -X $HTTP_METHOD -H "Authorization: Bearer $TOKEN" "https://api.digitalocean.com/v2/$OBJECT"
```
andrewsomething, line 1217:
```suggestion
        curl -X $HTTP_METHOD -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" "https://api.digitalocean.com/v2/$OBJECT"
```

Let's use `DIGITALOCEAN_TOKEN` to be consistent with the examples.

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1185,5 +1185,105 @@ components:
...
+        separated with a `&`. Tools like `curl` can create the appropriate URI when
+        given parameters and values; this can also be done using the `-F` flag and
+        then passing the key and value as an argument. The argument should take the
+        form of a quoted string with the attribute being set to a value with an
+        equal sign.
+
+        #### Pass Parameters as a JSON Object
+
+        ```
+            curl -H "Authorization: Bearer $TOKEN" \
```
andrewsomething, line 1244:
```suggestion
            curl -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" \
```

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1185,5 +1185,105 @@ components:
...
+                -H "Content-Type: application/json" \
+                -d '{"name": "example.com", "ip_address": "127.0.0.1"}' \
+                -X POST "https://api.digitalocean.com/v2/domains"
+        ```
+
+        #### Pass Filter Parameters as a Query String
+
+        ```
+             curl -H "Authorization: Bearer $TOKEN" \
```
andrewsomething, line 1253:
```suggestion
             curl -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" \
```

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1185,5 +1185,105 @@ components:
...
+        `Authorization` header containing your OAuth token. All requests must be
+        made over HTTPS.
+
+        #### Authenticate with a Bearer Authorization Header
+
+        ```
+        curl -X $HTTP_METHOD -H "Authorization: Bearer $TOKEN" "https://api.digitalocean.com/v2/$OBJECT"
+        ```
+
+        ### Parameters
```
andrewsomething, line 1220: Looks like this isn't a sub-topic of authentication in the current docs. It's it's own section:

https://developers.digitalocean.com/documentation/v2/#parameters

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1185,5 +1185,105 @@ components:
...
+
+        #### Pass Filter Parameters as a Query String
+
+        ```
+             curl -H "Authorization: Bearer $TOKEN" \
+                 -X GET \
+                 "https://api.digitalocean.com/v2/images?private=true"
+        ```
+
+        ### Cross Origin Resource Sharing
```
andrewsomething, line 1258: Same here:

https://developers.digitalocean.com/documentation/v2/#cors

andrewsomething (changes requested):

scotchneat (approved):
This is great @dbrian57! 

One question (not blocking because this is how it is in the current docs too): 
We describe how to use OAuth tokens for authentication but it's not specifically stated that one can use an API token as well. Do you know if that's intentionally left out?

dbrian57:
> I think it's just being a bit imprecise with the language and is using Oauth tokens and personal access tokens interchangeably.

@andrewsomething @scotchneat That's exactly the case and all of this could use a little tightening up, but I think that's outside the scope of this PR.

As for the requested changes, I've broken out each sub-section to mirror the sections on Developers. Thoughts?

scotchneat:
!deploy

scotchneat (dismissed):
LGTM!

specification/DigitalOcean-public.v2.yaml
dbrian57: Yeah, I tried the pull-right format and it didn't work. 🤷

andrewsomething (approved):

PR #513, opened 2021-04-26
registry: Fix validate name request body.

andrewsomething:
The wrong model was ref-ed for this operation. Additionally, the cURL example was missing the body altogether.


andrewsomething:
Apparently `models/validate_registry.yml` had just been sitting in my untracked files since I first added the registry endpoints... :facepalm:

scotchneat (approved):
lgtm

PR #514, opened 2021-04-26
dbaas: forking is a valid status.

andrewsomething:
When restoring from a backup or creating a replica, the status can be `forking`.


scotchneat (approved):
👍🏽

PR #515, opened 2021-04-26
droplet actions: use anyOf instead of allOf.

andrewsomething:
Both the rename and snapshot Droplet actions have the same attributes in their request body, `type` and `name`. This means we can not use `allOf` as that requires matching against _exactly one_ of the subschemas. Instead, we need `anyOf`.

https://github.com/digitalocean/openapi/blob/e7d843aba7629434683bea43889bbbe16563fe97/specification/resources/droplets/models/droplet_actions.yml#L70-L78

and

https://github.com/digitalocean/openapi/blob/e7d843aba7629434683bea43889bbbe16563fe97/specification/resources/droplets/models/droplet_actions.yml#L91-L99

Fixes: 

```
=== RUN   TestDropletAction_Rename/create_droplet_action-rename/contract
    prism_proxy.go:209: [CONTRACT TEST VIOLATION] code=oneOf message=should match exactly one schema in oneOf severity=Error location=request.body
```

### DIFF
--- a/specification/resources/droplets/post_droplet_action.yml
+++ b/specification/resources/droplets/post_droplet_action.yml
@@ -40,7 +40,7 @@ requestBody:
   content:
     application/json:
       schema:
-        oneOf:
+        anyOf:
           - $ref: 'models/droplet_actions.yml#/droplet_action_type'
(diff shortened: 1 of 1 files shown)


scotchneat (approved):
Nice find

PR #517, opened 2021-04-27
images: Updates to address contract validation errors.

andrewsomething:
This PR bundles three updates to address contract validation errors:

* Use anyOf for path attribute. This is a workaround for https://github.com/stoplightio/prism/issues/1339 Technically they are both strings as they are in a URL path, but retaining the underlaying type info is useful for documentation and clients. We use this for SSH keys as well: https://github.com/digitalocean/openapi/blob/4caff6ae7b96b5f0030bda19edfaf8e45effb58e/specification/resources/ssh_keys/parameters/ssh_key_identifier.yml#L6
* `min_disk_size` and `size_gigabytes` can be `null` in response to the initial POST to create a custom image as that info is not yet available.
* `status` can be both `NEW` or `new` and prism is case sensitive. A bit unfortunate, but we're documenting reality here not the ideal design.

### DIFF
--- a/specification/resources/images/get_image.yml
+++ b/specification/resources/images/get_image.yml
@@ -21,7 +21,7 @@ parameters:
       **Private** images *must* be identified by image `id`.
     required: true
     schema:
-      oneOf:
+      anyOf:
         - type: integer
         - type: string
     examples:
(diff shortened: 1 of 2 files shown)


scotchneat (approved):
Nice

PR #521, opened 2021-04-27
snapshots: Use anyOf for path attribute.

andrewsomething:
Like in https://github.com/digitalocean/openapi/pull/517, use `anyOf` for path attribute. This is a workaround for stoplightio/prism#1339 Technically they are both strings as they are in a URL path, but retaining the underlying type info is useful for documentation and clients.


scotchneat (approved):
lgtm

PR #522, opened 2021-04-28
Use anyOf for flip and image actions.

andrewsomething:
Resolves multiple instances of this contract violation error due to overlapping properties.

    code=oneOf message=should match exactly one schema in oneOf severity=Error location=request.body

### DIFF
--- a/specification/resources/floating_ips/models/floating_ip_actions.yml
+++ b/specification/resources/floating_ips/models/floating_ip_actions.yml
@@ -1,31 +1,34 @@
 floating_ip_action_type:
   type: object
   required:
-    - type
+  - type
   properties:
     type:
       type: string
+      enum:
+      - assign
+      - unassign
       description: The type of action to initiate for the floating IP.
   discriminator:
     propertyName: type
     mapping:
-      ASSIGN: '#/floating_ip_action_unassign'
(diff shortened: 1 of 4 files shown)


ChiefMateStarbuck (approved):

specification/resources/floating_ips/models/floating_ip_actions.yml
```diff
@@ -1,31 +1,34 @@
...
 
 floating_ip_action_assign:
   allOf:
     - $ref: '#/floating_ip_action_type'
     - type: object
       required:
-        - type
-        - droplet_id
+      - type
+      - droplet_id
```
scotchneat, line 31: This discriminator ref's to objects that ref back to `floating_ip_action_type`. Does this not cause any circular reference issues?

andrewsomething: Not sure. I don't remember the full context, but that was added to resolve an error w/ openapi generator.

https://github.com/digitalocean/openapi/pull/483

In this PR, I'm just flipping the bad logic here:

```
      ASSIGN: '#/floating_ip_action_unassign'
      UNASSIGN: '#/floating_ip_action_assign'
```

Could just leave that out as unrelated.

scotchneat (approved):
👍🏽

PR #523, opened 2021-04-28
dbaas: Use pattern not format when UUID can be blank.

andrewsomething:
Replaces `format: uuid` with a regex matching a UUID as well as an empty string. Fixes multiple instances of this contract viloation:

    code=format message=should match format \"uuid\" severity=Error location=request.body.private_network_uuid
    code=format message=should match format \"uuid\" severity=Error location=request.body.rules[0].uuid
    code=format message=should match format \"uuid\" severity=Error location=request.body.rules[0].cluster_uuid

### DIFF
--- a/specification/resources/databases/models/database_cluster.yml
+++ b/specification/resources/databases/models/database_cluster.yml
@@ -58,7 +58,7 @@ properties:
     readOnly: true
   private_network_uuid:
     type: string
-    format: uuid
+    pattern: '^$|[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}'
     example: d455e75d-4858-4eec-8c95-da2f0a5f93a7
     description: >-
       A string specifying the UUID of the VPC to which the database cluster will be
--- a/specification/resources/databases/models/firewall_rule.yml
+++ b/specification/resources/databases/models/firewall_rule.yml
@@ -3,12 +3,12 @@ type: object
 properties:
   uuid:
     type: string
-    format: uuid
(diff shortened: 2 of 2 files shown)


andrewsomething:
Setting `omitempty` in godo for these would also silence these errors in the tests. I'm a bit conflicted. If a blank string is valid, than the spec should likely reflect that. Though that may also be the case of many other  attributes were `omitempty` is set...

https://github.com/digitalocean/godo/blob/acba81d61dd9f11dacd03e3ca62a9596a09cfa88/databases.go#L210

https://github.com/digitalocean/godo/blob/acba81d61dd9f11dacd03e3ca62a9596a09cfa88/databases.go#L305-L306

specification/resources/databases/models/database_cluster.yml
```diff
@@ -58,7 +58,7 @@ properties:
     readOnly: true
   private_network_uuid:
     type: string
-    format: uuid
+    pattern: '^$|[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}'
```
scotchneat, line 61:
```suggestion
    pattern: '^(>$|[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12})?'
```

scotchneat (approved):
> Setting `omitempty` in godo for these would also silence these errors in the tests. I'm a bit conflicted. If a blank string is valid, than the spec should likely reflect that. Though that may also be the case of many other attributes were `omitempty` is set...

Ideally, the service shouldn't accept a blank string but, considering we're working with a production API, I do think it's worth looking into the impact of setting the `omitempty` tag on these fields. 

For the purpose of addressing the contract failures, I like this approach.

PR #524, opened 2021-04-28
load balancer: Use pattern not format when ip can be blank.

andrewsomething:
IP addresses are not available for a load balancer in the response while still creating. Resolves:

    code=format message=should match format \"ipv4\" severity=Error location=response.body.load_balancer.ip
    code=format message=should match format \"ipv4\" severity=Error location=response.body.load_balancers[1].ip


scotchneat (approved):
👍🏽

PR #529, opened 2021-05-24
Description Sub-headers

dbrian57:
The headers in the current docs are abnormally large. See here: https://docs.digitalocean.com/reference/api/api-reference/#section/Introduction/Curl-Examples

I've marked them down to an h3.


ChiefMateStarbuck (approved):
You're right, this does look weird. Good catch thank you. will push on green.

PR #535, opened 2021-07-27
droplet: Add with_droplet_agent attribute for create requests.

andrewsomething:


specification/resources/droplets/models/droplet_create.yml
```diff
@@ -93,6 +93,16 @@ properties:
...
 
+  with_droplet_agent:
+    type: boolean
+    example: true
+    description: A boolean indicating whether to install the DigitalOcean agent
+      used for providing access to the Droplet web console in the control panel.
+      By default, the agent is installed on new Droplets but installation errors
+      are ignored. To prevent it from being installed, set to `false`. To make
+      installation errors fatal, failing the create request, explicitly set it
+      to `true`.
```
andrewsomething, line 104: @house-lee Can you confirm that I am correctly describing the logic here?

house-lee: It looks good to me, but to be exact, it's only affecting how we react to incompatibility (i.e. OS not supported), we (as the droplet service) can't predict if there will be any installation error at this point. So not sure if you want to reveal that detail to customers, but I'll leave it up to you.

andrewsomething: :+1:  Added "OS not supported" as an example of the type of error we're talking about.

house-lee (commented):
👍

ChiefMateStarbuck (approved):

PR #536, opened 2021-07-28
domains: Fix example for creates and add ip_address attribute.

andrewsomething:
The example for creating a new domain actually shows the domain record payload. While fixing this, I noticed that `ip_address` attribute was also missing despite being mentioned in the description.


bentranter (approved):
LGTM!

PR #540, opened 2021-08-12
volumes: Document creating from a snapshot.

andrewsomething:
It was reported that we are missing the `snapshot_id` attribute required for creating a new volume from a snapshot. This was in the old docs and omitted here by oversight. I also realized that we are not labeling the attributes required for creates.


dbrian57 (approved):
LGTM! Thanks for taking this on. It was in my queue: PDOCS-1219

PR #544, opened 2021-09-17
app: Reorder list endpoint for better display.

andrewsomething:
This fixes PR fixes a pet peeve of mine. It reorders the Apps endpoints in the root spec file with the aim of making them display more logically in the side bar. It moves the core CRUD operations to the top of the list and groups related operations more closely.

Old order:

New order:


scotchneat (approved):
LGTM

PR #546, opened 2021-09-21
add api requests for interacting with app alerts

notxarb:
This adds documentation for the new API calls to manage alerts for the App Platform.
Creating/Updating/Removing of alerts is handled by the app specification.
Managing recipients of alerts is managed by the API to avoid having PII in the app spec.
There are two new API calls:
* List Alerts - This lists the alerts that are registered to an app.
* Update Alert Destinations - This allows users to update the list of recipients to alert messaging.


andrewsomething:
!deploy

andrewsomething:
!deploy

andrewsomething (approved):
:+1:  LGTM, thanks!

PR #548, opened 2021-09-21
Indicate Encoding of dbaas CA Certificate

rpmoore:
Updating the get ca description to indicate that the certificate response is base64 encoded


specification/resources/databases/get_ca.yml
```diff
@@ -7,7 +7,7 @@ description: |
   `/v2/databases/$DATABASE_ID/ca.
```
andrewsomething, line 7:
```suggestion
  `/v2/databases/$DATABASE_ID/ca`.
```

While we're in here. Looks like we're missing a closing backtick on this.

andrewsomething (approved):
LGTM!

PR #551, opened 2021-09-29
k8s: add updated docs for ha support

sunny-b:
This updates the K8s docs with support for the HA control plate and the `ha` field in the create and update commands.


bentranter (approved):
Looks great, thanks @sunny-b!

PR #556, opened 2021-10-15
resources/droplets: adds floating ips to associated resources docs

srbiv:
Floating IPs are now a supported resource. This adds them to the
examples and mentions them in the documentation.


andrewsomething:
!deploy

PR #561, opened 2021-11-01
monitoring metrics: Provide real-world example responses.

andrewsomething:
While working on https://github.com/digitalocean/godo/pull/491, I realized it would be very helpful if we provided some examples that are closer to real-world responses. While all the metrics responses use the same model, there is some variation in the responses by endpoint due to different labels being used and the type of data returned.


scotchneat (approved):
LGTM

PR #566, opened 2021-11-22
Add missing required field

lfundaro:
The API contract requires name to be
provided and this is currently missing
from the docs


andrewsomething:
!deploy

andrewsomething (approved):
:+1:  Thanks for catching that @lfundaro

PR #567, opened 2021-11-22
registry/CON-3906: Added documnentation for new registry/http endpoints: ListRepositoriesV2 and ListManifests

CollinShoop:
- Adds docs for repository `ListManifests` endpoint
- Adds docs for repository `ListRepositoriesV2` endpoint
- Deprecates `ListRepositories` endpoint

### DIFF
--- a/specification/DigitalOcean-public.v2.yaml
+++ b/specification/DigitalOcean-public.v2.yaml
@@ -1124,9 +1124,13 @@ paths:
     post:
       $ref: 'resources/registry/validate_registry_name.yml'
 
-  /v2/registry/{registry_name}:
+  /v2/registry/{registry_name}/repositories:
     get:
       $ref: 'resources/registry/list_registry_repositories.yml'
+
+  /v2/registry/{registry_name}/repositoriesV2:
+    get:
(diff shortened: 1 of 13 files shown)


andrewsomething:
!deploy

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1124,10 +1124,14 @@ paths:
     post:
       $ref: 'resources/registry/validate_registry_name.yml'
 
-  /v2/registry/{registry_name}:
+  /v2/registry/{registry_name}/repositories:
     get:
       $ref: 'resources/registry/list_registry_repositories.yml'
```
andrewsomething, line 1129: If we're going to be deprecating v1 of this endpoint, we'll want to add `deprecated: true` to  `resources/registry/list_registry_repositories.yml`

CollinShoop: Added, along with some additional language to prompt users to instead of the [V2] endpoint.

andrewsomething (approved):
:+1:  Looks great!

CollinShoop:
The http endpoints being documented are now available.

PR #568, opened 2021-11-22
Add k8s cluster as supported project resource

scotchneat:
Updating docs to reflect that kubernetes clusters can be added to projects.

### DIFF
--- a/specification/DigitalOcean-public.v2.yaml
+++ b/specification/DigitalOcean-public.v2.yaml
@@ -473,8 +473,8 @@ paths:
     post:
       $ref: 'resources/apps/cancel_deployment.yml'
 
-  ? /v2/apps/{app_id}/deployments/{deployment_id}/components/{component_name}/logs
-  : get:
+  /v2/apps/{app_id}/deployments/{deployment_id}/components/{component_name}/logs:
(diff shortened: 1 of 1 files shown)


specification/DigitalOcean-public.v2.yaml
```diff
@@ -472,8 +473,8 @@ paths:
     post:
       $ref: 'resources/apps/cancel_deployment.yml'
 
-  /v2/apps/{app_id}/deployments/{deployment_id}/components/{component_name}/logs:
-    get:
+  ? /v2/apps/{app_id}/deployments/{deployment_id}/components/{component_name}/logs
+  : get:
```
andrewsomething, line 477: Was this (and the other similar changes) intentional? It is valid YAML, but think it's probably less clear to most human readers

scotchneat: Definitely not a formatter issue (it would have affected all the paths). I must have fat-fingered it somehow.

andrewsomething (approved):
:+1:  Looks good, thanks for updating this!

ChiefMateStarbuck (approved):

PR #569, opened 2021-11-23
APPS-4420 - Preserve route and don't rewrite request

ElanHasson:


scotchneat:
!deploy

specification/resources/apps/models/app_route_spec.yml
```diff
@@ -6,3 +6,7 @@ properties:
       all components within an app.
     type: string
     example: /api
+  preserve_path_prefix:
+    description:  Boolean. An optional flag to preserve the path that is forwarded to the backend service. By default, the HTTP request path will be trimmed from the left when forwarded to the component. For example, a component with `path=/api` will have requests to `/api/list` trimmed to `/list`. If this value is `true`, the path will remain `/api/list`.
```
scotchneat, line 10: Does `Boolean` need to be specified in the description if the property has `type: boolean`?

```suggestion
    description: An optional flag to preserve the path that is forwarded to the backend service. By default, the HTTP request path will be trimmed from the left when forwarded to the component. For example, a component with `path=/api` will have requests to `/api/list` trimmed to `/list`. If this value is `true`, the path will remain `/api/list`.
```

ElanHasson: Sorry, just saw this now.

scotchneat (approved):
LGTM. Thanks @ElanHasson!

PR #571, opened 2021-12-02
Database config

rpmoore:


specification/resources/databases/responses/databases_copy.yml
```diff
@@ -0,0 +1,22 @@
+description: A JSON object with a key of `databases`.
```
bentranter, line 1: This file appears to be identical to [github.com/digitalocean/openapi/specification/resources/databases/responses/databases.yml](https://github.com/digitalocean/openapi/blob/3d352a60aa4404069eaf777619a157b74aa9b5b1/specification/resources/databases/responses/databases.yml), could we remove this one? It doesn't seem like the new endpoint references this response, so it might've been mistakenly added in the PR you picked up on.

specification/resources/databases/examples/curl/get_database_config.yml
```diff
@@ -0,0 +1,6 @@
+lang: cURL
+source: |-
+  curl -X POST \
```
bentranter, line 3: Should this be `GET` instead of `POST`?

specification/resources/databases/patch_database_config.yml
```diff
@@ -0,0 +1,47 @@
...
+  - Databases
+
+requestBody:
+  required: true
+  content:
+    application/json:
+      schema:
+        allOf:
+          - $ref: 'models/database_config.yml'
+          - type: object
```
bentranter, line 21: I believe that the `allOf` statement is unnecessary here, since you can use the reference directly:
```suggestion
  content:
    application/json:
      schema:
        $ref: ''models/database_config.yml''
```

bentranter (changes requested):
A few small tweaks, but looking good overall!

ChiefMateStarbuck:
!deploy

specification/resources/databases/examples/curl/get_database_config.yml
rpmoore: Good catching, fixing!

specification/resources/databases/responses/databases_copy.yml
rpmoore: yeah, they are identical. Not sure why it was added and I can't find any references to it. I'll go ahead and remove it.

specification/resources/databases/patch_database_config.yml
rpmoore: I think this is meant to be `oneOf`, I know the plan for this API call is for it to accept different payloads depending on the engine type. I do think that the `- type: object` needs to be removed though.

bentranter:
Looking at this again, I see that the response for `PATCH` isn't showing up as expected (see https://api-engineering.nyc3.digitaloceanspaces.com/spec-ci/redoc-index.html?pr=571#operation/patch_database_config).

@rpmoore, do you want me to take over this PR from here? I know it can be quite a bit of work to iterate on this given our tooling (and the pain of writing a ton of YAML 😛) – I'd be happy to take this over and then tag you for a final review of the generated docs to confirm that it's as expected.

bentranter:
Closing in favour of #590

PR #572, opened 2021-12-29
Elaborates on DBaaS slugs and the Windows CLI

yGabriel:
## Description and Details
Resolves PDOCS-1192 by elaborating on DBaaS slugs and adding a paragraph on following `curl` examples with the Windows CLI.

## Before approving, please test or review...
Verify technical wording and style.

## Checklist

- Site documentation and test coverage
    - [X] This PR does not require additional site documentation or test coverage, or
    - [ ] This PR includes necessary site documentation or test coverage.
- Blockers
    - [X] This PR can be merged as soon as it is approved, or
    - [ ] This PR is complete but should not be merged when it is approved; I have added the `DO-NOT-MERGE` label and described when it can be merged above, or
    - [ ] This PR is incomplete and should not be merged; I have added the `WIP` label and tagged a specific reviewer for an in-progress review.

### DIFF
--- a/specification/DigitalOcean-public.v2.yaml
+++ b/specification/DigitalOcean-public.v2.yaml
@@ -157,39 +157,39 @@ tags:
       which is used in some of the following requests. Each slug denotes the node's identifier,
       CPU count, and amount of RAM, in that order.
 
-      For Basic nodes, reference the following table for its slug:
+      For **Basic nodes**, reference the following table for its slug:
 
-      Slug               | Machine Type           | CPU     | RAM
-      -------------------|------------------------|---------|---------
-      db-s-1vcpu-1gb     | Basic node             | 1 vCPU  | 1 GB
-      db-s-1vcpu-2gb     | Basic node             | 1 vCPU  | 2 GB
-      db-s-2vcpu-4gb     | Basic node             | 2 vCPU  | 4 GB
-      db-s-4vcpu-8gb     | Basic node             | 4 vCPU  | 8 GB
-      db-s-6vcpu-16gb    | Basic node             | 6 vCPU  | 16 GB
-      db-s-8vcpu-32gb    | Basic node             | 8 vCPU  | 32 GB
-      db-s-16vcpu-64gb   | Basic node             | 16 vCPU | 64 GB
+      Slug               | CPU     | RAM
+      -------------------|---------|---------
+      db-s-1vcpu-1gb     | 1 vCPU  | 1 GB
+      db-s-1vcpu-2gb     | 1 vCPU  | 2 GB
+      db-s-2vcpu-4gb     | 2 vCPU  | 4 GB
+      db-s-4vcpu-8gb     | 4 vCPU  | 8 GB
+      db-s-6vcpu-16gb    | 6 vCPU  | 16 GB
+      db-s-8vcpu-32gb    | 8 vCPU  | 32 GB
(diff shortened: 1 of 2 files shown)


danaelhe:
looks great! Will merge after code freeze.

specification/DigitalOcean-public.v2.yaml
```diff
@@ -149,22 +149,37 @@ tags:
...
+
+      Slug Size    | RAM   | CPU     | Storage |
+      -------------|-------|---------|---------|
+      1vcpu-1gb    | 1 GB  | 1 vCPU  | 10 GB   |
+      1vcpu-2gb    | 2 GB  | 1 vCPU  | 25 GB   |
+      2vcpu-4gb    | 4 GB  | 2 vCPU  | 38 GB   |
+      4vcpu-8gb    | 8 GB  | 4 vCPU  | 115 GB  |
+      6vcpu-16gb   | 16 GB | 6 vCPU  | 270 GB  |
+      8vcpu-32gb   | 32 GB | 8 vCPU  | 580 GB  |
+      16vcpu-64gb  | 64 GB | 16 vCPU | 1.12 TB |
```
andrewsomething, line 178: Are all of these actually available in the `db-s`, `gd`, and `so1_5` variants?  For example, looking in the control panel, I see the smallest "General Purpose Node" available for me is 8 GB RAM and 2 vCPUs which doesn't match any of these. How many size slugs are available in total? Would it make sense to explicitly list them all?

yGabriel: Hi, Andrew! You're absolutely right; the node plans are different based on machine type. I also found that there are exceptions based on database engine. For example, MongoDB offers greater disk size (and offers standby nodes with the smallest node size/machine type), and Redis has a slightly different node plan description.

Without those exceptions, there are around 19 permutations to potentially list. I'll take a look at this today and notify you for another review.

andrewsomething:
!deploy

specification/description.yml
```diff
@@ -260,6 +260,12 @@ introduction: |
   Throughout this document, some example API requests will be given using the
   `curl` command. This will allow us to demonstrate the various endpoints in a
   simple, textual format.
+    
+  These examples assume that you are using a Linux or macOS command line. To run
+  these `curl` commands in a Windows command line, you must change all single
+  quotes (`''`) in the command to double quotes (`""`) and escape nested quotes
+  with a slash (`/`). For example, instead of `'{"size":"db-s-4vcpu-8gb", "num_nodes":3}'`,
+  use `"{\"size\":\"db-s-4vcpu-8gb\", \"num_nodes\":3}"`.
```
andrewsomething, line 268: I'm not on a Windows machine, but I'm curious if using PowerShell over the command prompt is an option here. I know it supports single quotes, but don't know if curl is available for it or not. If it is, that might be easier than having to rewrite the commands.

andrewsomething: This doesn't need to be blocking if you don't know off hand. We can always come back to this. Honestly, I feel like this whole curl examples section could use a bit of a rewrite.

yGabriel: Just finished following up on this! Have up-to-date solutions for cmd.exe, PowerShell, and WSL methods with the help of @ElanHasson.

yGabriel:
!deploy

andrewsomething:
!deploy

andrewsomething (approved):
:+1:  Thanks @yGabriel!

PR #574, opened 2022-01-13
Set proxy port to 8000.

andrewsomething:
At some point, the port used for the proxy changed from `8080` to `8090`. IIRC, this was so you can run the docs preview and the proxy at the same time. This broke some assumptions in the Makefile. This update makes it so the port used in the docker compose file and Makefile match. I chose `8000` as it's more memorable for me that `8090`, but don't really have a strong opinion.


bentranter (approved):
👍

PR #575, opened 2022-01-13
Update NPM dependencies.

andrewsomething:
Just doing a bit of housekeeping. I've validated that all the basics (bundle, lint, proxy) are working as expected after the upgrade.


danaelhe (approved):
thanks asb!

PR #579, opened 2022-01-26
Update LBaaS examples to use `size_unit` field

dbrian57:
What is says on the tin. We've deprecated the `size` field, so we should be using the `size_unit` field in the LBaaS docs.


specification/resources/load_balancers/examples/go/create_load_balancer.yml
```diff
@@ -15,7 +15,7 @@ source: |-
 
       createRequest := &godo.LoadBalancerRequest{
           Name:      "example-01",
-          Size:      "lb-medium",
+          Size_unit: "1",
```
andrewsomething, line 18:
```suggestion
          SizeUnit: "1",
```

This is actually `SizeUnit` in godo. In general, Go code uses Pascal Case 

https://github.com/digitalocean/godo/blob/4b7fcd34ac65b0b5e05e8c074f3263b9e99b01c5/load_balancers.go#L37

specification/resources/load_balancers/examples/go/update_load_balancer.yml
```diff
@@ -15,7 +15,7 @@ source: |-
 
       updateRequest := &godo.LoadBalancerRequest{
           Name:      "example-01",
-          Size:      "lb-medium",
+          Size_unit: "2",
```
andrewsomething, line 18:
```suggestion
          SizeUnit: "2",
```

andrewsomething (changes requested):
Thanks @dbrian57! I left a suggested change inline for the Go examples.

dbrian57:
@andrewsomething  Oh right! Good catch. Thank you. Committed your changes.

andrewsomething (approved):

PR #580, opened 2022-01-26
Updated Smallest VPC Size Available

dbrian57:
We now support `/28` sized networks. Tiny change for tiny networks.


andrewsomething (approved):
:shipit:

PR #581, opened 2022-01-27
Fix create volume snapshot code example

bentranter:
Relates to https://github.com/digitalocean/godo/issues/512

The current example for creating a volume snapshot incorrectly lists the example as accepting only a volume ID as an argument, but godo requires that a struct be passed (see https://pkg.go.dev/github.com/digitalocean/godo#StorageServiceOp.CreateSnapshot).


andrewsomething (approved):

andrewsomething:
Thanks Ben!

PR #595, opened 2022-03-11
adding "volume_name" to example in block storage

danaelhe:
the doc and curl example has "volume_name" while the payload did not.

https://docs.digitalocean.com/reference/api/api-reference/#operation/post_volume_action_by_name


ChiefMateStarbuck (approved):
Thank you @danaelhe

PR #597, opened 2022-03-22
added empty_content

ChiefMateStarbuck:
droplet 204's have a content-header set in the response.


scotchneat:
It would also make sense to add the content-type expected to the spec.

Just keep in mind that this will likely have to be done to more endpoints.

scotchneat (dismissed):
The change looks good. 
I would prefer naming `empty_content.yml` as`no_content_with_content_type.yml` since it closer represents the `204 No Content` HTTP response code. But this is not a blocker.

scotchneat (approved):
👍🏽

PR #602, opened 2022-04-05
Updated Volume Limit

dbrian57:
Users can attach up to seven volumes to a Droplet, not five.


andrewsomething (approved):
:+1: good on green

PR #606, opened 2022-04-07
Removes required field

scotchneat:
In the case of our admin test account that doesn't get billed, we can see the `billing_history` field isn't included in the response.

The example got auto-formatted also.


bentranter (approved):
@scotchneat LGTM!

PR #607, opened 2022-04-07
Remove type: string from kubconfig response

scotchneat:
Since the content type is `application/yaml`, the body is technically an object (default type).
This should resolve the current contract test violation for `TestKubernetes_CreateBasic/get_kubernetes_cluster_kubeconfig`:

```
12:46:18     round_trippers.go:313: [CONTRACT TEST VIOLATION] response body doesn't match the schema: Field must be set to string or not be present
12:46:18         Schema:
12:46:18           {
12:46:18             "type": "string"
12:46:18           }
12:46:18         
12:46:18         Value:
12:46:18           "object"
```

### DIFF
--- a/specification/resources/kubernetes/responses/kubeconfig.yml
+++ b/specification/resources/kubernetes/responses/kubeconfig.yml
@@ -10,9 +10,6 @@ headers:
 
 content:
   application/yaml:
-    schema:
-      type: string
-
     example: |
       apiVersion: v1
       clusters:


andrewsomething (approved):
:+1:  lgtm assuming the linter is happy with it (Can you have an example response for something without a response schema?)

scotchneat:
> Can you have an example response for something without a response schema?

I tested it locally and the test now passes with this change, so I'd say yes.

scotchneat:
Not sure why that linter's taking so long. It finishes in about 42s locally.

andrewsomething:
I've noticed that recently as well in other PRs here recently. It's not providing live logs, so it's hard to tell what is going on.

PR #612, opened 2022-04-26
monitoring: alerts for Load Balancers TLS conns/s utilization

anitgandhi:
Internal ticket number: LBAAS-2200


scotchneat:
!deploy

scotchneat (approved):
We could move that enum to a separate file so it can be ref'd and only updated in one place but that's not blocking for this change.

LGTM

PR #616, opened 2022-05-03
droplets operationID changes

ChiefMateStarbuck:


andrewsomething (commented):
As this is the first of the renames, we should establish the pattern we want to use across all of them and think about the resulting method names. 

- `Droplet` or `Droplets`?  Think we should go with the plural. That matches the HTTP routes (e.g. `/v2/droplets`) and godo (`client.Droplets.Create()`).

- We should probably look to use the shortest, clear version of the name. Only one here that jumps out to me is `droplets_list_all` could just be `droplets_list`.

scotchneat:
In addition to what Andrew suggested, it's a good idea to add details about the motivation for a change to the description. While there are internal details documented, this serves as a good place for the community to understand the reasoning.

I'd also be curious to see if this change had the intended effect with client generation.

danaelhe:
> As this is the first of the renames, we should establish the pattern we want to use across all of them and think about the resulting method names.
> 
> * `Droplet` or `Droplets`?  Think we should go with the plural. That matches the HTTP routes (e.g. `/v2/droplets`) and godo (`client.Droplets.Create()`).
> * We should probably look to use the shortest, clear version of the name. Only one here that jumps out to me is `droplets_list_all` could just be `droplets_list`.

Also, use camelcase for resources have underscores.

ChiefMateStarbuck:
> In addition to what Andrew suggested, it's a good idea to add details about the motivation for a change to the description. While there are internal details documented, this serves as a good place for the community to understand the reasoning.
> 
> I'd also be curious to see if this change had the intended effect with client generation.

Great Idea. this would be good for the main <- operationID PR

PR #617, opened 2022-05-04
OperationID fixes: 1clicks

danaelhe:
Changing naming for 1clicks for a smoother Autorest ingestion:

- Changing preface of 1_clicks to 1clicks so Autorest can see it as one entity. 
- Changed references to it anywhere in the spec for consistency.

created a bundle of the spec with the new changes and didn't appear to run into any reference issues.


danaelhe:
Ah man, Sam's commits somehow got put into this PR

ChiefMateStarbuck:
We should be pushing to `OperationID` here instead of `main`

danaelhe:
> We should be pushing to `OperationID` here instead of `main`

ah dang, i thought i was

PR #621, opened 2022-05-05
Revert1 clicks changes

danaelhe:
Accidentally pushed changes straight to OperationID without reviewing. Reverting so I can create another PR that will have a review for 1clicks.


ChiefMateStarbuck (approved):

ChiefMateStarbuck (commented):
I like the changes you've made so this PR might not be necessary

PR #634, opened 2022-05-06
regions: Update operationId for new pattern.

andrewsomething:
Updates the `operationId` for the new `{resource}_{action}` pattern


danaelhe (approved):
thanks!

PR #641, opened 2022-05-09
registry: Update operationId for new pattern.

andrewsomething:
Updates the `operationId` for the new `{resource}_{action}` pattern. It retains the use of `registry` (singular) in order to be consistent with the path (`/v2/registry`).


specification/DigitalOcean-public.v2.yaml
```diff
@@ -1134,71 +1134,71 @@ paths:
...
+      $ref: 'resources/registry/registry_get_subscription.yml'
 
     post:
-      $ref: 'resources/registry/post_registry_subscription.yml'
+      $ref: 'resources/registry/registry_update_subscription.yml'
 
   /v2/registry/docker-credentials:
     get:
-      $ref: 'resources/registry/get_docker_credentials.yml'
+      $ref: 'resources/registry/registry_get_docker_credentials.yml'
```
ChiefMateStarbuck, line 1154: `registry_get_dockerCredentials`

endpoint_action_everythingElseInCamel

This applies to a few endpoints

ChiefMateStarbuck (dismissed):
Thank you for helping out @andrewsomething

specification/DigitalOcean-public.v2.yaml
andrewsomething: Is that what we're doing elsewhere?  Feels a bit odd to me to mix camel case and snake case this way. 

For grouping, autorest requires the resource to be the first fragment before the first underscore (`{resource}_{action}`). Is it opinionated about what comes after that? How does it effect the generated client?

danaelhe: Yeah, it is weird but autorest can get funny with multiple underscores *sometimes*. I think it was with volume_actions where it would drop some words from the generated python functions signatures after the second underscore. So we tried to keep it to just one underscore. So with volumeActions_create_byName you'll see that autorest created a function that's "create_by_name", so that confirms they understood how to handle the camel case for byName. 

I did just run autorest with these changes and it seems like autorest did create a "build_get_docker_credentials_request". So in this case, it seems fine. Maybe it's when the resource has an underscore that trips autorest up.

danaelhe (changes requested):
The operationID in `specification/resources/registry/registry_list_repository_manifests.yml ` needs to be updated: registry_list_repository_manifests

danaelhe (approved):
lgtm, thanks!

PR #649, opened 2022-05-12
Fix pg version on the create database examples

shabtaisharon:
pg 10 is no longer available therefore I've updated all the example to use the latest pg version


ChiefMateStarbuck (approved):

ChiefMateStarbuck:
!deploy

PR #656, opened 2022-05-18
APPS-5636 App Platform updates

ElanHasson:
* Add Functions Alert rule types
* Spelling and grammar fixes
* Update minimumTLSVersion to only support 1.2 and 1.3


specification/resources/apps/models/app_domain_spec.yml
```diff
@@ -40,6 +40,13 @@ properties:
     type: string
     format: hostname
     example: example.com
-
+  minimumTLSVersion:
```
kamaln7, line 43:
```suggestion
  minimum_tls_version:
```

kamaln7 (approved):

rcj4747 (approved):

rcj4747 (dismissed):

specification/resources/apps/models/app_domain_spec.yml
```diff
@@ -40,6 +40,13 @@ properties:
     type: string
     format: hostname
     example: example.com
-
+  minimum_tls_version:
+    type: string
+    format: string
```
andrewsomething, line 45:
```suggestion
```

We've actually recently removed all other instances of  `format: string` from this as some OpenAPI tooling didn't understand how to handle it. Technically it should be fine as the spec describes `format` as an open string-valued property, but since `string` as a format is redundant with the type, we took the simple path of removing it.  

https://swagger.io/specification/#data-types

andrewsomething (changes requested):
Looks good, but one small change requested inline.

andrewsomething (approved):

PR #657, opened 2022-05-20
APICLI-1284: Add Trigger Python Client Generation Workflow

danaelhe:
This workflow will trigger the workflow in [digitalocean/digitalocean-client-python](https://github.com/digitalocean/digitalocean-client-python/pull/5) to re-generate the client when a change is done the spec.

### DIFF
--- a/.github/workflows/trigger-python-client-gen.yml
+++ b/.github/workflows/trigger-python-client-gen.yml
@@ -11,6 +11,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: trigger-workflow
-        run: gh workflow run --repo digitalocean/digitalocean-client-python python-client-gen.yml --ref main
+        run: gh workflow run --repo digitalocean/digitalocean-client-python python-client-gen.yml --ref main -f commit_url=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/commit/$GITHUB_SHA
         env:
           GITHUB_TOKEN: ${{ secrets.OPENAPI_SECRET }}
--- a/.github/workflows/trigger-python-client-gen.yml
+++ b/.github/workflows/trigger-python-client-gen.yml
@@ -5,7 +5,7 @@ on:
       - 'specification/**'
 jobs:
   build:
-    name: Re-generate Python Client
+    name: Trigger digitalocean-client-python Workflow
     runs-on: ubuntu-latest
     steps:
       - uses: convictional/trigger-workflow-and-wait@v1.3.0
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-python-client-gen.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  build:
+    name: Trigger digitalocean-client-python Workflow
+    runs-on: ubuntu-latest
+    steps:
+      - uses: convictional/trigger-workflow-and-wait@v1.3.0
+        with:
+          owner: digitalocean
+          repo: digitalocean-client-python
+          github_token: ${{ secrets.OPENAPI_SECRET }}
```
danaelhe, line 15: secret was created in this repo. token is stored in lastpass.

.github/workflows/trigger-python-client-gen.yml
```diff
@@ -0,0 +1,16 @@
+name: Trigger Python Client Generation
+on: 
+  push:
+    paths:
```
andrewsomething, line 4:
```suggestion
on: 
  push:
    branches:
      - main
    paths:
```

andrewsomething: Think we only want this to trigger for pushes to main (or the `operationID` feature branch for now).

.github/workflows/trigger-python-client-gen.yml
```diff
@@ -0,0 +1,16 @@
...
+      - 'specification/**'
+jobs:
+  build:
+    name: Trigger digitalocean-client-python Workflow
+    runs-on: ubuntu-latest
+    steps:
+      - uses: convictional/trigger-workflow-and-wait@v1.3.0
```
andrewsomething, line 11: Looks like this blocks until the workflow completes in the other repo. Wonder if we need all its features? Always a bit wary of third-party actions having access to secrets. Surprised that GitHub doesn't offer something official for this. We could just hit the API endpoint.

I think the GitHub CLI is available on the runners. Using that, it should be something like:

    gh workflow run --repo digitalocean/digitalocean-client-python python-client-gen.yml --ref main

https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event

danaelhe: that makes sense! I was able to replace it out seamlessly. thanks for the suggestion.

andrewsomething (approved):
:+1:  LGTM!

andrewsomething (approved):

scotchneat (approved):
👍🏽

PR #658, opened 2022-05-23
Bump rules to error

ChiefMateStarbuck:
This PR bumps the new rules to the level `error` and removes the old rules.

Changes in droplets were found while updating the rules.


scotchneat (approved):
lgtm

PR #662, opened 2022-05-24
fix bug for 0 value of integer

ChiefMateStarbuck:
previously, 0 was not a valid example for an `integer`.

This PR fixes this bug


scotchneat (approved):
👍🏽

PR #664, opened 2022-06-02
remove extra "new" in enum

danaelhe:
causing client gen failures...not sure how it got added and if it's needed


scotchneat (approved):
It looks like the actual value returned for status on a create is `NEW` so this LGTM.

PR #669, opened 2022-06-13
Volumecreate

ChiefMateStarbuck:
Per suggestions by the team, volumeactions now has a default function entitled `attach` and a secondary function entitled `attach_byId`


danaelhe:
Looks great! So looks like the default "volume_actions.attach()" will be "attach by name" and the alternative way to attach will be "volume_actions.attach_by_id()" 

I'm ok with that, unless someone thinks they should be switched.

scotchneat (approved):
lgtm

PR #676, opened 2022-06-23
Fixes quotes in app domain spec

scotchneat:
Submitting this on its own so it's not blocked by the discussion about model changes in #670.

This change makes the field and enum since only the two listed strings are allowed and I was able to leave the double quotes within back-ticks for it to successfully render in preview like this:

image

I think the text specifying "Must be one of the following values ..." is superfluous after making the field an enum but I'll leave that for discussion.

I also generated the client from this change and was able to successfully import it and use it to create a droplet.

I'll also push a commit to #670 to remove this from that PR.

### DIFF
--- a/specification/resources/apps/models/app_domain_spec.yml
+++ b/specification/resources/apps/models/app_domain_spec.yml
@@ -44,8 +44,10 @@ properties:
     type: string
     maxLength: 3
     minLength: 3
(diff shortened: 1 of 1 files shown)


ChiefMateStarbuck (approved):
dont worry about the failed spec validation. That's already been fixed in pushmain

PR #681, opened 2022-06-27
Change volumeActions_attach to volumeActions_run

scotchneat:
`attach` is one of the enum values for the `type` field of volumeActions so these file/operation names are incorrect.

`run` is already an allowed alias for `POST` in the spectral rules so this should work.

I have some other suggestions that I'll create a separate PR for.

### DIFF
--- a/specification/DigitalOcean-public.v2.yaml
+++ b/specification/DigitalOcean-public.v2.yaml
@@ -1363,7 +1363,7 @@ paths:
 
   /v2/volumes/actions:
     post:
-      $ref: 'resources/volumes/volumeActions_attach.yml'
+      $ref: 'resources/volumes/volumeActions_run.yml'
 
   /v2/volumes/snapshot/{snapshot_id}:
     get:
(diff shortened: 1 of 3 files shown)


specification/DigitalOcean-public.v2.yaml
```diff
@@ -1363,7 +1363,7 @@ paths:
 
   /v2/volumes/actions:
     post:
-      $ref: 'resources/volumes/volumeActions_attach.yml'
+      $ref: 'resources/volumes/volumeActions_run.yml'
```
andrewsomething, line 1366: Looks like we are using `post` for Droplet actions. No strong preference, but I think what ever we choose should be consistent across both.

https://github.com/digitalocean/openapi/blob/80f4892ed175e5d797452dfc95a551bcf8866b1f/specification/resources/droplets/dropletActions_post.yml#L1

andrewsomething (approved):

scotchneat:
> I think we ended up with attach here based on my feedback. There had been volume_actions.create_by_name() and volume_actions.create_by_id(). Those are OK, but could be better. A user doesn't really want to "create" an action. They want to attach or detach a volume, so attach() and detach() might be more friendly. 

I see. I wasn't aware of the previous discussion. My thought here is that when the operation is named "attach" would result in a generated client where detaching a volume action would look like this: `client.volume_actions.attach(111, {"type":"detach", "droplet_id":222})`. 

I was thinking it'd make more sense to be consistent with the naming as you mention in the other comment and add patch methods to the `VolumeActions` class with more intuitive functions (e.g. `client.volumes_actions.attach(<volume_id>, <droplet_id>) / client.volume_actions.detach(<volume_id>, <droplet_id>)`) in the generated client.

PR #683, opened 2022-06-28
Clarify operations are on rules

scotchneat:
e.g. We're not listing firewalls, we're listing the firewall rules.

This is just a nitpick I've had for a while so I figured I'd put this out here now that we're renaming operations. The main reason for this is that when I tested the generated client as is, the method names suggest we're listing/deleting/getting firewalls (when really we're dealing with firewall rules). 

(I realize the url is still `.../firewall`)


andrewsomething (approved):
:+1:  Good call

PR #684, opened 2022-07-07
Update token used to trigger Python client workflow.

andrewsomething:
The API token being used to trigger Python client workflow does not have the correct permissions leading to the workflow failing with:

```
HTTP 401: Bad credentials (https://api.github.com/repos/digitalocean/digitalocean-client-python/actions/workflows/python-client-gen.yml)
Try authenticating with:  gh auth login
```

https://github.com/digitalocean/openapi/runs/7222091456?check_suite_focus=true#step:2:7

This new secret should have the correct permissions.


scotchneat (approved):
LGTM

PR #685, opened 2022-07-07
Don't eat new lines in tag descriptions.

andrewsomething:
The yaml scalar being used in many of the tag descriptions strips out line breaks. This updates it to use the literal style. 

https://yaml.org/spec/1.2.2/#23-scalars


andrewsomething:
!deploy

PR #691, opened 2022-07-15
Upgrade spectral dependency

scotchneat:
Upgrading `spectral` required significant changes to spectral configs and functions so I'm only including relevant changes to the upgrade in this PR.

I used the following migration guide:
https://meta.stoplight.io/docs/spectral/ZG9jOjg2MDIwMDM-spectral-v5-to-v6-migration-guide

A couple of things to call out are:
- The `--skip-rule` cli flag was removed so I had to create a separate `bundled.spectral.yml` to use for the second command in `make lint` which lints the bundled spec.
- The `operation-default-response` core function was removed. Since all we did was disable the rule, I removed it from the config.
- The function definitions required updating - https://meta.stoplight.io/docs/spectral/ZG9jOjI1MTkw-custom-functions#writing-functions

for APICLI-1407

### DIFF
--- a/spectral/ruleset.yml
+++ b/spectral/ruleset.yml
@@ -1,22 +1,34 @@
+aliases:
+  PathItem:
+    - $.paths[*]
+  OperationObject:
+    - '#PathItem[get,put,post,delete,options,head,patch,trace]'
+
+functions:
+  - ensurePropertiesExample
+  - ensureAllArraysHaveItemTypes
+  - ensureSnakeCaseWithDigits
+  - validateOpIDNaming
+
 rules:
   ratelimit-headers:
     description: Response must include ratelimit-x headers
-    message: "{{description}}; missing {{property}}"
+    message: '{{description}}; missing {{property}}'
     severity: error
     given: $..responses.*
     then:
-    - field: headers.ratelimit-limit
-      function: truthy
-    - field: headers.ratelimit-remaining
-      function: truthy
-    - field: headers.ratelimit-reset
-      function: truthy
+      - field: headers.ratelimit-limit
+        function: truthy
+      - field: headers.ratelimit-remaining
+        function: truthy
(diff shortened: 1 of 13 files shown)


scotchneat:
`49s`
Boom! You lookin for this?

spectral/ruleset.yml
```diff
@@ -1,30 +1,42 @@
+aliases:
+  PathItem:
+    - $.paths[*]
+  OperationObject:
+    - '#PathItem[get,put,post,delete,options,head,patch,trace]'
```
scotchneat, line 5: I noticed they added a way to alias JSONPaths.

spectral/ruleset.yml
```diff
@@ -74,72 +86,62 @@ rules:
...
 
   common-responses-too-many-requests:
     description: Responses should contain common response - 429 (too many requests)
-    message: "{{description}}. Missing {{property}}"
+    message: '{{description}}. Missing {{property}}'
     severity: error
     given: $.paths..responses
     then:
-      - field: '429'
-        function: truthy
```
scotchneat, line 100: This was leftover from troubleshooting. It'll work either way and since there's only one item listed, the new change makes sense anyway.

spectral/ruleset.yml
```diff
@@ -74,72 +86,62 @@ rules:
...
-    message: "{{description}}; {{value}} incorrect"
-    then:
-      function: ensureSnakeCaseWithDigits
-
-
   operationid-must-follow-new-naming-conventions:
     description: operationIds must follow naming conventions for method
     type: style
-    given: "$.paths[*][*]"
+    given: '#OperationObject'
```
scotchneat, line 142: Example of referencing an alias

ChiefMateStarbuck (approved):
🏆

PR #692, opened 2022-07-18
Update remaining node dependencies

scotchneat:
The main remaining dependency was:

* `@redocly/openapi-cli` was deprecated and renamed to `@redocly/cli`. Also upgraded to: `^1.0.0-beta.59`

The following were also updated. The `test` make target does exercise the newman mock server and run mocked tests using the collection generated. It might still be an interesting toolset but we don't frequently use either in our workflow anymore so we could consider removing them.

* Upgraded `newman` to `^5.1.2`
* Upgraded `openapi-to-postmanv2` to `4.0.0`

I also removed the `--only=dev` flag from the `npm install` command as the `dev` option has been deprecated. The default behavior now is to install all dependencies. When the flag is passed `--only=production`, the command only installs production dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ help:
 
 .PHONY: dev-dependencies
 dev-dependencies: ## Install development tooling using npm
-	npm install --only=dev
+	npm install
 
 .PHONY: start-mockedproxy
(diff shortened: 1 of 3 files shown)


andrewsomething (approved):
:+1:

PR #697, opened 2022-07-25
billing/pdf: add content type header

bentranter:
The response for `/v2/customers/my/invoices/{uuid}/pdf` contains a `content-type` header in the response, so we should include that in the spec.

### DIFF
--- a/specification/resources/billing/responses/invoice_pdf.yml
+++ b/specification/resources/billing/responses/invoice_pdf.yml
@@ -1,7 +1,9 @@
-description: >- 
+description: >-
   The response will be a PDF file.
 
 headers:
+  content-type:
+    $ref: '../../../shared/headers.yml#/content-type'
   content-disposition:
     $ref: '../../../shared/headers.yml#/content-disposition'
   ratelimit-limit:
@@ -12,7 +14,7 @@ headers:
     $ref: '../../../shared/headers.yml#/ratelimit-reset'
 
 content:
-  application/pdf: 
+  application/pdf:
     schema:
       type: string
       format: binary


specification/resources/billing/responses/invoice_pdf.yml
```diff
@@ -12,7 +14,7 @@ headers:
     $ref: '../../../shared/headers.yml#/ratelimit-reset'
 
 content:
-  application/pdf: 
+  application/pdf:
```
scotchneat, line 17: I thought this is how the content type was determined but, you're right, the content-type isn't added to the rendered docs without it being set in the headers.

specification/resources/billing/responses/invoice_pdf.yml
```diff
@@ -1,7 +1,9 @@
-description: >- 
+description: >-
   The response will be a PDF file.
 
 headers:
+  content-type:
+    $ref: '../../../shared/headers.yml#/content-type'
```
scotchneat, line 6: Unfortunately, this ref defines the content-type as `application/json`

https://github.com/digitalocean/openapi/blob/main/specification/shared/headers.yml#L51

You might need to add a new schema for pdf content type.

scotchneat (changes requested):
I'm not sure this will have the intended effect.
If you run `make preview` with this change, it shows up like this:

image

specification/resources/billing/responses/invoice_pdf.yml
andrewsomething: That's shown here, but doesn't document that the header is actually included"

image

specification/resources/billing/responses/invoice_pdf.yml
andrewsomething: Ya... It's the "example."  It's really just documenting that the header is returned, not what it's content type is. Looks like we have the same problem with the content-disposition header example as well. It uses `csv`

PR #700, opened 2022-07-27
Address issues found during model generation.

scotchneat:
Includes:

`app_domain_spec.yml`: The `domain` property has `format` and `pattern` values.  These are conflicting during model generation as they both represent a string formatting
`droplet.yml`: The `kernel` property is required and the $ref'd schema has `nullable: true`. The resulting model requires the field and causes validation errors when a value isn't supplied for the property.
`forwarding_rules`: Multiple schemas have this property with the incorrect `minimum` schema property set.

APICLI-1463

### DIFF
--- a/specification/resources/apps/models/app_domain_spec.yml
+++ b/specification/resources/apps/models/app_domain_spec.yml
@@ -2,7 +2,6 @@ type: object
 properties:
   domain:
     type: string
-    format: hostname
     maxLength: 253
     minLength: 4
     pattern: ^((xn--)?[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*\.)+[a-zA-Z]{2,}\.?$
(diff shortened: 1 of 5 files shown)


ChiefMateStarbuck (approved):

ChiefMateStarbuck:
!deploy

andrewsomething (commented):
!deploy

danaelhe:
Does this fix the weird namings for the following models that were edited?

andrewsomething (commented):
!deploy

scotchneat:
> Does this fix the weird namings for the following models that were edited?

@danaelhe no, that will require moving schemas that are defined inline (like those for request and response bodies) into separate schema files. The model generator tool also has an option (`--use-title-as-name`) that we could use to ensure generated models use a specific name. Although, it didn't work all that great and I didn't spend much time digging into it for this change.

andrewsomething (approved):
:+1:  This all looks useful even if we don't go with the model generation tool.

PR #708, opened 2022-08-03
Removes unnecessary `allOf` for eviction policy

scotchneat:
The current definition works but causes errors during client generation. 

image

It seems the refs to the eviction_policy_model was wrapped in unnecessary `allOf`. This simplifies the refs and fixes the generation errors.

I also added a job to the pr workflow that generates a client using the spec from the PR changes.

### DIFF
--- a/.github/workflows/spec.pr.yml
+++ b/.github/workflows/spec.pr.yml
@@ -12,7 +12,6 @@ on:
       - 'spectral/**'
 
 jobs:
-
   lint:
     name: Validate spec
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


scotchneat:
😮‍💨

scotchneat:
Apologies for the force-push noise. [nektos/act ](https://github.com/nektos/act) doesn't run the jobs I added to the workflow so I couldn't test this locally before pushing.

I realized it would make more sense for the workflow changes to be their own PR but this worked out because it allowed me to verify the errors were fixed within the same change.

ChiefMateStarbuck (approved):
!deploy

PR #711, opened 2022-08-09
Add support for /databases/options GET

DWizGuy58:
**What**

- Adds documentation on `/databases/option` endpoint which returns regions, versions, slugs, etc that are available for the various engines we support.

**Why**

- Provides details to clients who would like to know the available options for databases

**Evidence**

image

### DIFF
--- a/specification/resources/databases/models/database_layout_options.yml
+++ b/specification/resources/databases/models/database_layout_options.yml
@@ -1,0 +1,8 @@
+type: object
+
+properties:
+  layouts:
+    type: array
+    readOnly: true
+    items:
+      $ref: './database_layout_option.yml'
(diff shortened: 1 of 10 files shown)


specification/resources/databases/models/database_layout_options.yml
```diff
@@ -0,0 +1,8 @@
+type: object
+
+properties:
+  layouts:
+    type: array
+    readOnly: true
```
andrewsomething, line 6:
```suggestion
    readOnly: true
    description: Layouts describe....
```

Would be great to have a description here. I'm not sure "layouts" are a well understood concept for users.

andrewsomething (commented):
Overall looks great. I did have one suggestion inline.

PR #715, opened 2022-09-07
serverless Namespaces public APIs

v-shant:
[RFC](https://docs.google.com/document/d/1WZW6admF1QpAJIag9E0lnLoxQFxWlyUEMX0D5qWxuE8/edit#)


bentranter (changes requested):
Hey @v-shant, a few high level pieces of feedback:

* The operation ID names need some changes. We use the operation IDs to generate code, so they need to be very consistent. The pattern is roughly `<namespace>_<crud-operation>`, so for example, `create_namespace` should instead be `namespaces_create`. This allows us to generate code like `client.namespaces.create()`.
* The filenames should match the operation IDs for consistency. For example, `get_namespace.yml` should be `namespaces_get.yml`.

If you could fix those, I'll go over this PR in more detail. We'll also need to wait for someone from the product docs team to go over this as well before merging (hence the label that was added).

andrewsomething:
> The operation ID names need some changes. We use the operation IDs to generate code, so they need to be very consistent. The pattern is roughly <namespace>_<crud-operation>, so for example, create_namespace should instead be namespaces_create. This allows us to generate code like client.namespaces.create().

Do we want `namespaces` to be the top level here with methods to look like `client.namespaces.create()`? I think `serverless` or `functions` would be the better option. So the operation ID `serverless_create_namespace` in that case, generating `client.serverless.create_namespace()`

specification/DigitalOcean-public.v2.yaml
```diff
@@ -469,7 +469,10 @@ tags:
       setups or other configurations requiring movable addresses.
 
       Reserved IPs are bound to a specific region.
-
+  
+  - name: Serverless
```
piyush0101, line 473: this should be `Functions`?

v-shant:
> > The operation ID names need some changes. We use the operation IDs to generate code, so they need to be very consistent. The pattern is roughly _, so for example, create_namespace should instead be namespaces_create. This allows us to generate code like client.namespaces.create().
> 
> Do we want `namespaces` to be the top level here with methods that look like `client.namespaces.create()`? I think `serverless` or `functions` would be the better option. So the operation ID would be `serverless_create_namespace` in that case, generating `client.serverless.create_namespace()`

can't use functions_create_namespace as its warning me 

```
operationid-must-follow-new-naming-conventions  POST /v2/functions/sandbox - functions_get_sandbox should end with one of: create, post, add, tag, install, reset, upgrade, recycle, run, retry, validate, assign, unassign, cancel, destroy, delete, update, attach, revert, commit. Prefer 'create' for new resources. Custom verbs may be acceptable for clarity. Example OperationID: droplet_create
```

bentranter:
> Do we want `namespaces` to be the top level here with methods that look like `client.namespaces.create()`? I think `serverless` or `functions` would be the better option. So the operation ID would be `serverless_create_namespace` in that case, generating `client.serverless.create_namespace()`

Yeah good call, I wasn't thinking about the design of this at a high enough level.

From Piyush's comment, it seems like `functions` may be the best name (looks like that change is already being made).

I also wonder about the create sandbox endpoint. It seems like this is a find-or-create type of endpoint. @andrewsomething do you have any thoughts about how something like this should be named? Maybe just find_or_create like you see in ORMs?

specification/resources/serverless/sandbox_get.yml
```diff
@@ -0,0 +1,33 @@
+operationId: sandbox_get
+
+summary: Sandbox
+
+description: >-
+  Creates or Gets existing sandbox namespace for the given uuid.
```
dbrian57: Per the conversation in Slack, we're not documenting this call, right?

specification/resources/serverless/responses/list_namespaces.yml
```diff
@@ -0,0 +1,22 @@
+description: The response will be an array of JSON objects with a key called `namespaces`.
```
dbrian57:
```suggestion
description: |- 
  An array of JSON objects with a key called `namespaces`.  Each object represents a namespace and contains
  the properties associated with it. 
```

specification/resources/serverless/responses/list_namespaces.yml
```diff
@@ -0,0 +1,22 @@
...
+  This will be set to an array of namespace objects, each of which will
```
dbrian57:
```suggestion

```

specification/resources/serverless/responses/list_namespaces.yml
```diff
@@ -0,0 +1,22 @@
...
+  contain the standard attributes associated with a namespace.
```
dbrian57:
```suggestion

```

piyush0101:
> > Do we want `namespaces` to be the top level here with methods that look like `client.namespaces.create()`? I think `serverless` or `functions` would be the better option. So the operation ID would be `serverless_create_namespace` in that case, generating `client.serverless.create_namespace()`
> 
> Yeah good call, I wasn't thinking about the design of this at a high enough level.
> 
> From Piyush's comment, it seems like `functions` may be the best name (looks like that change is already being made).
> 
> I also wonder about the create sandbox endpoint. It seems like this is a find-or-create type of endpoint. @andrewsomething do you have any thoughts about how something like this should be named? Maybe just find_or_create like you see in ORMs?

I discussed with Vishant, we'll be removing the `sandbox` endpoint from the spec. It powers the now-deprecated sandbox functionality via `doctl`. This endpoint is not intended for use by the end users, we'll eventually remove it.

dbrian57 (changes requested):
This is looking good. The documentation updates are inline. 

I still don't have a clear understanding of the `api-host` field, so that will need to be cleared up to. I tried to be thorough about the line formatting/indentation but it might need another pass after the feedback has been implemented.

Let me know if you have any questions. I'd like to do one more pass after this round. Thanks, Vishant!

specification/resources/serverless/models/create_namespace.yml
```diff
@@ -0,0 +1,15 @@
+type: object
+properties:
+  region:
+    type: string
+    example: nyc
+    description: Region in which namespace will be created.
+
+  label:
+    type: string
+    example: my namespace
```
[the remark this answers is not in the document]
v-shant: this is freehand text that user can speciify

specification/resources/serverless/models/namespace_info.yml
```diff
@@ -0,0 +1,32 @@
...
+properties:
+  api_host:
+    type: string
+    example: https://api_host.io
+    description: The api host for the namespace.
+
+  namespace:
+    type: string
+    example: fn-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+    description: A unique string format of uuid with a prefix fn-.
```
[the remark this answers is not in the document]
v-shant: No, both are different text

specification/resources/serverless/sandbox_get.yml
v-shant: yes

v-shant:
> I still don't have a clear understanding of the `api-host` field, so that will need to be cleared up to. I tried to be thorough about the line formatting/indentation but it might need another pass after the feedback has been implemented.

api_host basically is a endpoint URL for the APIs

specification/resources/functions/models/not_allowed.yml
```diff
@@ -0,0 +1,21 @@
...
+
+  messages:
+    description: A list of error messages.
+    type: object
+    properties:
+      '':
+        type: array
+        items: 
+          type: string
+        example: ['not allowed to get namespace']
```
andrewsomething, line 17: Why do 403 errors have a custom format that doesn't match other error responses? Looking at the version history of the RFC, looks like this is a recent addition.

v-shant: from some RPCs we have to return sone validation errors, like if namespace not present then the error should be Namespace not found (404), similar to this we have a condition where user is not allowed to get response for few namespaces (not restricted to all namespaces). in that case we are returning codes.PermitionDenied and error message= not allowed

on the aggregator side this is being converted to in that format

andrewsomething: The error code isn't the issue, it is the format of the response body. We document an error format that is used across the API. You 404 error uses it, but not your 403.

```
    {
      "id":       "forbidden",
      "message":  "not allowed to get namespace",
      "request_id": "88d17b7a-630b-4083-99ce-5b91045efdb4"
    }
```

v-shant: my bad, I don't know from where I got this. just verify in postman and its 

```
{
    "id": "forbidden",
    "message": "not allowed to get namespace",
    "request_id": "b11e45a4-892c-48c9-9001-b6cffe9fe795"
}
```

updating the doc and RFC

andrewsomething: Thanks!

specification/resources/functions/examples/curl/create_namespace.yml
```diff
@@ -0,0 +1,7 @@
+lang: cURL
+source: |-
+  curl -X POST \
+    -H "Content-Type: application/json" \
+    -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" \
+    -d '{"region": "nyc", "label": "my namespace label"}' \
```
andrewsomething, line 6:
```suggestion
    -d '{"region": "nyc1", "label": "my namespace label"}' \
```

specification/resources/functions/models/create_namespace.yml
```diff
@@ -0,0 +1,15 @@
+type: object
+properties:
+  region:
+    type: string
+    example: nyc
```
andrewsomething, line 5:
```suggestion
    example: nyc1
```

specification/resources/functions/models/namespace_info.yml
```diff
@@ -0,0 +1,34 @@
+type: object
+
+properties:
+  api_host:
+    type: string
+    example: https://api_host.io
+    description: The api host for the namespace.
```
dbrian57, line 7:
```suggestion
    description: |- 
      The namespace's API hostname. Each function in a namespace is provided an endpoint at the namespace's hostname.
```

dbrian57 (approved):
Minor edits, but looks good otherwise 👍

specification/resources/functions/models/namespace_info.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  namespace:
```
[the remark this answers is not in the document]
v-shant: no, in the API response key is namespace

bentranter (dismissed):
I'm happy with this now, but let's wait for the other reviewers to re-review before merging

specification/resources/functions/models/namespace_info.yml
```diff
@@ -0,0 +1,35 @@
+type: object
+
+properties:
```
scotchneat, line 3: The actual api call includes `created_at` and `updated_at` fields in the response. 
Unless there's a reason we don't want those documented, I think they should be included here.

specification/resources/functions/models/create_namespace.yml
```diff
@@ -0,0 +1,15 @@
+type: object
+properties:
+  region:
+    type: string
+    example: nyc1
+    description: The [datacenter region](https://docs.digitalocean.com/products/platform/availability-matrix/#available-datacenters) in which to create the namespace.
```
scotchneat, line 6: Since functions aren't available in all regions, we should make sure the linked region reference includes a `Functions Availability` section. Otherwise, the available regions should be listed here.

specification/resources/functions/functions_delete_namespace.yml
```diff
@@ -0,0 +1,44 @@
+operationId: functions_delete_namespace
+
+summary: Delete Namespace
+
+description: >-
+  Deletes the given namespace.  When a namespace is deleted all assets, in the namespace are deleted, this
+  includes packages, functions and triggers. Deleting a namespace is a destructive operation and assets in the
+  namespace are not recoverable after deletion. Some metadata is retained, such as activations, or soft deleted for reporting purposes.
+
+  To delete namespace, send a DELETE request to `/v2/functions/namespaces/{{namespace_id}}`.
```
scotchneat, line 10: The rest of the spec uses the following pattern in docs. It would be good to keep this consistent.

```suggestion
  To delete namespace, send a DELETE request to `/v2/functions/namespaces/$NAMESPACE_ID`.
```

specification/resources/functions/functions_get_namespace.yml
```diff
@@ -0,0 +1,42 @@
+operationId: functions_get_namespace
+
+summary: Get Namespace
+
+description: >-
+  Gets the namespace details for the given namespace UUID.
+  To get namespace details, send a GET request to `/v2/functions/namespaces/{{namespace_id}}` with no parameters.
```
scotchneat, line 7: Same as above.

```suggestion
  To get namespace details, send a GET request to `/v2/functions/namespaces/$NAMESPACE_ID` with no parameters.
```

scotchneat (changes requested):
Overall, the change looks great to me. 

I did bring up a few points and added some suggestions in-line.

specification/resources/functions/models/create_namespace.yml
rabbah: @scotchneat are you asking that we link to the available regions only? there is a page in pdocs that says where functions are available that's region based so NYC rather than NYC1 or NYC3.

availability is here https://docs.digitalocean.com/products/functions/details/availability/ although i just noticed this is not up to date.

specification/resources/functions/models/namespace_info.yml
v-shant: how about removing them, they are just timestamp stored in the database at the time of adding namespace,

specification/resources/functions/models/create_namespace.yml
piyush0101: Functions are available in the following regions - ams3, blr1, fra1, lon1, nyc1, sfo3, sgp1, tor1

piyush0101 (dismissed):

specification/resources/functions/models/create_namespace.yml
scotchneat: > are you asking that we link to the available regions only? 

@rabbah Right. It's a bit misleading as is. Since neither these API docs nor the referenced availability matrix indicate there is limited availability.

scotchneat (approved):
I think the availability matrix documentation can be handled on the pdocs side at this point so it doesn't have to block this PR. (See my [comment in the review thread](https://github.com/digitalocean/openapi/pull/715#discussion_r972216955)).

Everything else looks good 👍🏽

andrewsomething (approved):

PR #728, opened 2022-10-13
Move generate and push client logic to openapi workflow

danaelhe:
Before, we had an openapi/workflow that triggered a digitalocean-python-client/workflow to 1) generate the clients 2) run the mocked tests. With this change, it will look more like this: 
- When a PR is created in openapi spec files it will run pydo's mocked tests.
- When the openapi PR is pushed to main, it will generate the client and create PR for pydo repo.

This had to be done because the github action we were using could not trigger other workflows using GITHUB_TOKEN which was causing PR checks to choke on the digitalocena-python-client side:

https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs


scotchneat:
> o.O These checks shouldn't even run because the changes aren't in specifications/

I've noticed workflows aren't honoring any `paths` of `paths-ignore` values lately.

PR #729, opened 2022-10-19
Add Release Support and Rework Client Generation Trigger Workflow

danaelhe:
Adding releases to this repo to associate versions of the DO spec to versions of the generated Python Client. 

This change does the following to support releases:
- Is consistent with doctl's releasing pattern of following [semantic versioning](https://semver.org/)
- Adds supporting make commands: `make changes` and `make tags`
- Adds the bundled spec as an artifact to the release to be picked up by the python client 

This change does the following to support triggering the workflow to regenerate the python client:
- Triggers the Python Client repo only on digitalocean/openapi releases 

The workflow is depicted below:
DO Openapi Spec to Pydo Workflow (3)


.github/workflows/trigger-python-client-gen.yml
```diff
@@ -11,7 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event.workflow_run.conclusion == 'success' }}
     steps:
+      - name: "Determine tag"
+        run: echo "TAG_RELEASE=`curl  "https://api.github.com/repos/danaelhe/openapi/tags" | jq -r '.[0].name'`" >> $GITHUB_ENV
       - name: trigger-workflow
-        run: gh workflow run --repo digitalocean/digitalocean-client-python python-client-gen.yml --ref main -f commit_url=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/commit/$GITHUB_SHA
+        run: gh workflow run --repo danaelhe/digitalocean-client-python python-client-gen.yml --ref main -f release_version=$TAG_RELEASE
```
halkeye, line 17: *gasp*
you are the owner of this now?

scripts/changelog.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+tfile=$(mktemp /tmp/openapi-CHANGELOG-XXXXXX)
+github-changelog-generator -org danaelhe -repo openapi >"$tfile"
```
halkeye, line 6: no idea if your fork is intentional or not

```suggestion
github-changelog-generator -org digitalocean -repo openapi >"$tfile"
```

.github/workflows/trigger-python-client-gen.yml
danaelhe: GREAT catch! That was from testing 😆

PR #730, opened 2022-10-20
Add Release Support and Rework Client Generation Trigger Workflow

danaelhe:
Adding releases to this repo to associate versions of the DO spec to versions of the generated Python Client.

This change does the following to support releases:
- Is consistent with doctl's releasing pattern of following [semantic versioning](https://semver.org/)
- Adds supporting make commands: make changes and make tags
- Adds the bundled spec as an artifact to the release to be picked up by the python client

This change does the following to support triggering the workflow to regenerate the python client:
- Triggers the Python Client repo only on digitalocean/openapi releases

The workflow is depicted below:
DO Openapi Spec to Pydo Workflow (2)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,6 @@
 name: tagged-release
-
+# this workflow occurs on every release
+# it uploads the bundled spec as an artifact of the release
 on:
   push:
     tags:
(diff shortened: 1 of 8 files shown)


danaelhe:
This is strange that the github checks are trying to run when these changes aren't in the specification/ path -___-

danaelhe:
Yayyyy the checks are working!

PR #731, opened 2022-10-20
Remove 'on/pull_request/paths:' in workflow

danaelhe:
Learned in Github Actions Office Hours that it is a known bug, that github has documented [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks), for PRs that change files outside of paths specified in a workflow to choke the gh actions checks. 

Suggested we remove it since it's not an expensive check

also made minor change to spec to have this PR work lol

### DIFF
--- a/.github/workflows/spec.pr.yml
+++ b/.github/workflows/spec.pr.yml
@@ -5,11 +5,7 @@ name: Spec PR
 # The name of the uploaded file will be prefixed with the PR number.
 # e.g. `spec-ci/previews/123-spec.yaml
 
-on:
-  pull_request:
-    paths:
-      - 'specification/**'
-      - 'spectral/**'
+on: [push, pull_request]
 
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


.github/workflows/spec.pr.yml
```diff
@@ -5,11 +5,7 @@ name: Spec PR
 # The name of the uploaded file will be prefixed with the PR number.
 # e.g. `spec-ci/previews/123-spec.yaml
 
-on:
-  pull_request:
-    paths:
-      - 'specification/**'
-      - 'spectral/**'
+on: [push, pull_request]
```
andrewsomething, line 8:
```suggestion
on: [pull_request]
```

I don't think we want `push` here as we have a different workflow for that.

danaelhe: Oh I see, thanks! Just updated.

andrewsomething (approved):
:+1: LGTM

PR #733, opened 2022-10-26
apps: update create app docs and the app example

ghostlandr:
We just added project_id as a parameter to the CreateApp RPC, so this PR updates the documentation to reflect that. project_id also comes back on the newly created App so that's documented now too.

### DIFF
--- a/specification/resources/apps/models/app.yml
+++ b/specification/resources/apps/models/app.yml
@@ -52,6 +52,11 @@ properties:
     title: The ID of the account to which the application belongs
     type: string
     example: 4f6c71e2-1e90-4762-9fee-6cc4a0a9f2cf
+  project_id:
+    readOnly: true
+    type: string
+    example: 88b72d1a-b78a-4d9f-9090-b53c4399073f
+    title: The id of the project for the app. This will be empty if there is a fleet (project) lookup failure.
   region:
     $ref: apps_region.yml
   spec:
--- a/specification/resources/apps/models/apps_create_app_request.yml
+++ b/specification/resources/apps/models/apps_create_app_request.yml
@@ -1,6 +1,9 @@
 properties:
   spec:
     $ref: app_spec.yml
+  project_id:
+    type: string
+    description: The UUID of the project the app should be assigned.
 required:
 - spec
 type: object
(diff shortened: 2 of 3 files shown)


specification/resources/apps/models/app.yml
```diff
@@ -52,6 +52,11 @@ properties:
     title: The ID of the account to which the application belongs
     type: string
     example: 4f6c71e2-1e90-4762-9fee-6cc4a0a9f2cf
+  project_id:
+    readOnly: true
+    type: string
+    example: 88b72d1a-b78a-4d9f-9090-b53c4399073f
+    title: The id of the project for the app. This will be empty if there is a fleet (project) lookup failure.
```
kamaln7, line 59:
```suggestion
    title: The id of the project for the app. This will be empty if there is a lookup failure.
```

kamaln7 (dismissed):

specification/resources/apps/models/app.yml
```diff
@@ -52,6 +52,11 @@ properties:
     title: The ID of the account to which the application belongs
     type: string
     example: 4f6c71e2-1e90-4762-9fee-6cc4a0a9f2cf
+  project_id:
+    readOnly: true
+    type: string
+    example: 88b72d1a-b78a-4d9f-9090-b53c4399073f
+    title: The id of the project for the app. This will be empty if there is a lookup failure.
```
andrewsomething, line 59:
```suggestion
    title: The ID of the project the app is assigned to. This will be empty if there is a lookup failure.
```

specification/resources/apps/models/apps_create_app_request.yml
```diff
@@ -1,6 +1,9 @@
 properties:
   spec:
     $ref: app_spec.yml
+  project_id:
+    type: string
+    description: The UUID of the project the app should be assigned.
```
andrewsomething, line 6:
```suggestion
    description: The ID of the project the app should be assigned to. If omitted, it will be assigned to your default project.
```

We should describe the behavior if it is omitted from the request. I assume it would be assigned to your default project?

andrewsomething (commented):
Thanks for the update! I left a couple small suggestions in line.

specification/resources/apps/models/apps_create_app_request.yml
ghostlandr: That's a great change, thank you.

andrewsomething (dismissed):

ghostlandr:
New in this branch, the `with_projects` param here
image

ChiefMateStarbuck (approved):

PR #732, opened 2022-10-24
Apps API now supports Content-Type and Accept headers.

honeybadgerinc:

PR #738, opened 2022-11-02
Fix client gen: Add schema to App yaml responses

scotchneat:
Client generation has been failing with the following error:

```
warning | Modeler/MissingType | The schema 'region-features' has no type or format information whatsoever. Location:
   file:///Users/cgarza/dev/external/digitalocean-client-python/DigitalOcean-public.v2.yaml#/components/schemas/region-features
warning | Modeler/MissingType | The schema 'region-sizes' has no type or format information whatsoever. Location:
   file:///Users/cgarza/dev/external/digitalocean-client-python/DigitalOcean-public.v2.yaml#/components/schemas/region-sizes
fatal   | TypeError: Cannot read properties of undefined (reading 'type')
fatal   | Process() cancelled due to failure
error   |   Error: Plugin modelerfour reported failure.
error   | Autorest completed with an error. If you think the error message is unclear, or is a bug, please declare an issues at https://github.com/Azure/autorest/issues with the error message you are seeing.
make: *** [generate] Error 1
```

I was able to track it down to the changes in #732 adding the `application/yaml` content type responses in the App create request. I believe autorest requires a type for any `application/*` content.

Unfortunately, adding the schema triggers the `oas3-valid-media-example` rule to error with `"example" property type must be object ` for:  
* `components.responses.new_app.content.text/yaml.example` 
* `paths./v2/apps.post.requestBody.content.application/yaml.example`

Also, running `make lint` has been inconsistent when checking the ref'd spec. Sometimes it will fail with the following and others, without changing anything, it wont.

```
/Users/cgarza/dev/external/openapi/specification/resources/apps/apps_create.yml
 41:16  error  oas3-valid-media-example  "example" property type must be object  requestBody.content.application/yaml.example
```

### DIFF
--- a/specification/resources/apps/apps_create.yml
+++ b/specification/resources/apps/apps_create.yml
@@ -2,11 +2,13 @@ operationId: apps_create
 
 summary: Create a New App
 
-description: Create a new app by submitting an app specification. For documentation
-  on app specifications (`AppSpec` objects), please refer to [the product documentation](https://docs.digitalocean.com/products/app-platform/reference/app-spec/).
+description:
+  Create a new app by submitting an app specification. For documentation
+  on app specifications (`AppSpec` objects), please refer to [the product
+  documentation](https://docs.digitalocean.com/products/app-platform/reference/app-spec/).
 
 tags:
-- Apps
+  - Apps
 
 parameters:
   - $ref: parameters.yml#/accept
@@ -34,6 +36,8 @@ requestBody:
               routes:
                 - path: /api
     application/yaml:
+      schema:
+        $ref: models/apps_create_app_request.yml
       example: |
         spec:
           name: web-app
@@ -53,16 +57,16 @@ requestBody:
   required: true
 
 responses:
-  "200":
+  '200':
     $ref: responses/new_app.yml
 
-  "401":
+  '401':
     $ref: ../../shared/responses/unauthorized.yml
 
-  "429":
-    $ref: "../../shared/responses/too_many_requests.yml"
+  '429':
+    $ref: '../../shared/responses/too_many_requests.yml'
 
-  "500":
+  '500':
     $ref: ../../shared/responses/server_error.yml
 
   default:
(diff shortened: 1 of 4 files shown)


scotchneat:
I've started working on overriding the `oas3-valid-media-example` rule for the specific files/paths known to have `application/yaml`, but I'm not able to get it to work consistently. 

I'll include the rule changes so others can test them.

danaelhe:
> I've started working on overriding the `oas3-valid-media-example` rule for the specific files/paths known to have `application/yaml`, but I'm not able to get it to work consistently.
> 
> I'll include the rule changes so others can test them.

Thanks Cesar for tracking this down. That's so so weird it's not working consistently. I'd love to try it on my end.

scotchneat:
> I wonder if that's something worth making a spectral rule for future endpoints- "if your endpoint has application/yaml then throw a warning to add the path to oas3-valid-media-example paths to override" or something

That's what I'm sorting out now.

danaelhe:
Trying out ASB's suggestion in slack: 

> Think we just need to remove the | here:
> https://github.com/digitalocean/openapi/blob/1320e32509eb8f08a8d1a745cf9e9e3c7adfdd34/specification/resources/apps/apps_create.yml#L41

Does seem to remove the error, but he did notice lots of warning level stuff on the uptime check endpoints...
>   19277:9  warning  operationid-must-follow-new-naming-conventions  GET /v2/uptime/checks - uptime_checks_list should end with one of: get, list. Prefer 'get' for retrieving a single object and 'list' for multiple objects. Example OperationID: droplet_get

Not sure what's going on there... It does end in list and if the check is broken, why is it only complaining about uptime?

scotchneat:
I had tried removing the `|` after `example: ` and the preview renders the example like json.

scotchneat:
Per a discussion in slack, we agreed that removing the `application/yaml` content from the request is the best course of action. It allows us to generate the client and we won't have to add validation for this scenario. Also, we decided it would be enough documentation for users to see the header parameters documenting the endpoint accepts yaml content.

PR #739, opened 2022-11-02
Fix client generation errors

scotchneat:
Removes the `applciation/yaml` content type from the app create request added in #732.

For consistency, also removes the `applciation/yaml` content type from the new app response.


danaelhe (approved):
LGTM 👍

PR #740, opened 2022-11-02
Fix: failed Generate Main preview workflow

scotchneat:
The latest run of the `spec.main.yml` workflow [failed to publish the bundled spec](https://github.com/digitalocean/openapi/actions/runs/3381370539/jobs/5615454389) because it wasn't able to evaluate the `SPACES_PATH` and `SPACES_ENDPOINT` secrets. It turns out these were referenced as `env.SPACES...`. 
This config hasn't changed in quite some time so I'm not sure why it stopped working but this change hopefully resolves it.


andrewsomething (approved):
:+1:  Why not

PR #734, opened 2022-11-01
Add "Upload Pydo Spec" in Makefile

danaelhe:
Reworked the workflow a bit to upload a specific bundled spec tied to a commit for the Pydo workflow to pick up. 
Workflow is as follows: 
DO Openapi Spec to Pydo Workflow (5)

Sister PR to this one: https://github.com/digitalocean/digitalocean-client-python/pull/103

PR #742, opened 2022-11-03
Add env to upload step in workflow

scotchneat:
🤦🏽 I completely missed that the `env` details were shifted and not copied when reviewing #734.


danaelhe (approved):
good catch!

ChiefMateStarbuck (approved):

PR #737, opened 2022-11-01
Add project_id field to LBaaS Docs, removed required tag Storage

dbrian57:
Resolved [PDOCS-1676](https://jira.internal.digitalocean.com/browse/PDOCS-1676) and PDOCS-1655. I've added the `project_id` field to the LBaaS docs and have removed the required tag from the Block Storage `filesystem_type` attribute.

PR #744, opened 2022-11-04
volumes: filesystem_type is not required.

andrewsomething:
As pointed out in https://github.com/digitalocean/openapi/pull/737 `filesystem_type` is not actually required for creating a volume. As is, making it not required causes issues with client generation due to our use of a discriminator:

```
error   | DiscriminatorNotRequired | Semantic violation: Discriminator must be a required property. (components > schemas > volumewrite_file_systemtype)
  discriminator: {
  propertyName: 'filesystem_type',
  mapping: {
    EXT4: '#/components/schemas/volumes_ext4',
    XFS: '#/components/schemas/volumes_xfs'
  }
}
```

The discriminator is only being used to change the character constrains for `filesystem_label`. An XFS volume can have a 16 character label while an EXT4 once can only have 12 characters. We can simplify this and just relay on `anyOf` for the same purpose. It does change the display, but allows for an accurate discription of the API.

Current:

image

Proposed change:

image

### DIFF
--- a/specification/resources/volumes/models/attributes.yml
+++ b/specification/resources/volumes/models/attributes.yml
@@ -1,7 +1,5 @@
 volume_write_file_system_type:
   type: object
-  required:
-    - filesystem_type
   properties:
     filesystem_type:
       type: string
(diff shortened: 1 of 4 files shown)


bentranter (approved):

danaelhe (approved):
approve on all green checks 👍

PR #748, opened 2022-11-07
Install Poetry for client generation workflow.

andrewsomething:
With https://github.com/digitalocean/pydo/pull/114, the generation workflow now calls `poetry`, but we don't install it here. This leads to the workflow failing:

https://github.com/digitalocean/openapi/actions/runs/3413655909/jobs/5680646383#step:5:615


danaelhe (approved):
🤦

PR #752, opened 2022-11-15
ratelimit headers now lowercase

dbrian57:
Resolves [PDOCS-1715](https://jira.internal.digitalocean.com/browse/PDOCS-1715). After receiving feedback, I've changed the `ratelimit` headers to lowercase in the intro doc as to align with the actual HTTP responses the API returns.


scotchneat (approved):
LGTM

PR #753, opened 2022-11-16
kubernetes: add high availability flag to cluster update

timoreimann:
Also clarify that the property cannot be disabled once enabled.


danaelhe (commented):
!deploy

scotchneat (approved):

PR #769, opened 2023-01-05
updates to pg adv conf

DMW2151:
Updates DBAAS Postgres advanced config with warning on parameters that cannot be reduced.


andrewsomething (approved):
:+1:  LGTM!

PR #770, opened 2023-01-12
intro: link directly to tokens page to avoid redirect.

andrewsomething:
Just happened to notice that we link to the old URL for the tokens page in the control panel. This updates it to avoid the redirect.


scotchneat (approved):
Good on green

PR #774, opened 2023-01-24
Add dbaas version availability

dweinshenker:
**What**

- Adds `version_availability` to `/v2/databases/options` endpoint
- Adds `version_end_of_life` and `version_end_of_availability` to `/v2/databases` List, `v2/databases` Create, `v2/databases/{uuid}` Get endpoints

**Why**

`version_availability` exposes end-of-life (EOL) and end-of-availability (EOA) for specific engine/version pairs.

EOL: the final date at which the engine/version will no longer be supported
- if this is empty, there is no EOL set

EOA: the final date at which new clusters can be created with this version
- if this is empty, there is no EOA set


danaelhe:
Thanks for the PR! @dweinshenker could you update this branch?

dweinshenker:
> 

@danaelhe updated!

danaelhe (approved):

PR #776, opened 2023-01-31
apps: Add missing domains list to the app objects

rcj4747:
The list of domains on the app object was missing from the specification.  This PR updates the app and apps response examples as well as the app model.  The example responses show both an active and configuring domain.


specification/resources/apps/responses/examples.yml
```diff
@@ -80,8 +90,80 @@ apps:
...
+        live_url_base: https://sample-php.example.com
+        live_domain: sample-php.example.com
+        domains:
+        - id: 0831f444-a1a7-11ed-828c-ef59494480b5
+          phase: "ACTIVE"
+          spec:
+            domain: "sample-php.example.com"
+            type: "PRIMARY"
+            zone: "example.com"
+            minimum_tls_version: "1.3"
```
scotchneat, line 102: Can we add `validations` to these examples?
```suggestion
            minimum_tls_version: "1.3"
          validations:
            - txt_name: _acme-challenge.app.example.com
              txt_value: lXLOcN6cPv0nproViNcUHcahD9TrIPlNgdwesj0pYpk
```

specification/resources/apps/responses/examples.yml
```diff
@@ -202,6 +292,79 @@ app:
...
       live_domain: sample-golang-zyhgn.ondigitalocean.app
       project_id: 88b72d1a-b78a-4d9f-9090-b53c4399073f
+      domains:
+      - id: e206c64e-a1a3-11ed-9e6e-9b7b6dc9a52b
+        phase: "CONFIGURING"
+        spec:
+          domain: "sample-golang.example.com"
+          type: "PRIMARY"
+          zone: "example.com"
+          minimum_tls_version: "1.3"
```
scotchneat, line 302: Here as well?
```suggestion
          minimum_tls_version: "1.3"
          validations:
            - txt_name: _acme-challenge.app.example.com
              txt_value: lXLOcN6cPv0nproViNcUHcahD9TrIPlNgdwesj0pYpk
```

specification/resources/apps/responses/examples.yml
rcj4747: I'd be happy to, but first a question for you on best practices. I left out a these in the example because they're fields that won't be valid combined with other fields in the example.  In this case a managed domain (the `zone` field is specified) wouldn't ever have TXT validations to display.  Do we want the entire example to be valid or would we prefer examples of the fields even if they might not all go together?

scotchneat: That makes sense. Examples should be entirely valid. 
It's possible to create separate examples for a different response schema. Check out the droplet create response examples: 
https://github.com/digitalocean/openapi/blob/main/specification/resources/droplets/droplets_create.yml#L43-L47

In fact, if fields are mutually exclusive, they should be defined as such in different request schemas. The droplet create has a simple example of this as well: 
https://github.com/digitalocean/openapi/blob/main/specification/resources/droplets/droplets_create.yml#L39-L41

rcj4747: The fields aren't mutually exclusive in the same way as the droplet example, but the `validations` field in the response is currently only ever populated by the back-end when issuing/renewing an SSL cert of a user-managed domain, not a DO-managed domain.  Whether this is DO or user managed comes down to the user's spec in the request having a `zone` field in the domain spec; and the spec has so many combinations of options that we've gone with something representative but not exhaustive.
In our example request the user app spec defines a domain and includes a value in the `zone` field indicating this is DO-managed; I did that because this field is much more common to see than the validations field.  A discussion in slack focused on the less strict treatment of the app spec and concluded that the validations field could be omitted to keep things accurate and not exhaustive.

scotchneat: Let's disregard my suggestion as it's safe to keep the app spec model(s) basic. 

It would make more sense to add a note to the spec field description property that the API definition for it is kept basic and point users to the product docs for a more thorough description of the app spec.

scotchneat (approved):
lgtm

PR #777, opened 2023-02-01
Document operationId naming convention

scotchneat:
This change adds documentation about the naming convention followed for operationIds. I also updated the validation function to describe the error better and reference this documentation.

Additionally, I updated the operations that failed the validation. This might need to be discussed further. We might also need to update any documentation that links to the API docs with the original operationIds.


scotchneat:
We'd have to update that failed pydo test after merging this.

PR #782, opened 2023-02-08
Ensure PR status is posted for failed Python client generation.

andrewsomething:
I happened to notice that if the "Validate Client Generation" fails, the status is never updated and left pending, e.g. in this PR https://github.com/digitalocean/openapi/pull/777.  `if: always()` is needed on the final step.


ChiefMateStarbuck (approved):
Cool find 🔮

PR #783, opened 2023-02-14
Add semantic_version to cluster list/get responses

dweinshenker:
Support was added for including `semantic_version` in the cluster response


mikeberezowski (approved):
LGTM. Thanks 👍

PR #784, opened 2023-02-24
App Create CURL Example Updated

dbrian57:
Updated the App Create CURL example with a data field to show users how they can pass a spec to the API.


bentranter (approved):
LGTM!

dbrian57:
@bentranter Thank you 🙏

PR #807, opened 2023-06-07
Add readonly field for MongoDB

zbarahal-do:
Support for creating MongoDB read only users - merge is dependent on feature flag enablement.

### DIFF
--- a/specification/resources/databases/databases_add_user.yml
+++ b/specification/resources/databases/databases_add_user.yml
@@ -38,6 +38,11 @@ requestBody:
             mysql_settings:
               auth_plugin: mysql_native_password
 
+        Add New Read Only User (MongoDB Only):
+          value:
+            name: my-readonly
+            readonly: true
+
 responses:
   '201':
     $ref: 'responses/user.yml'


dbrian57 (approved):
@andrewsomething  Hey, this looks fine to me, but is the `Add New Read Only User (MongoDB Only)` line actually gonna show up in the docs? The `Add New User with MySQL Auth Plugin` doesn't show up as far as I can tell. Just worried about making it confusing.

ChiefMateStarbuck (approved):

danaelhe:
> @andrewsomething Hey, this looks fine to me, but is the `Add New Read Only User (MongoDB Only)` line actually gonna show up in the docs? The `Add New User with MySQL Auth Plugin` doesn't show up as far as I can tell. Just worried about making it confusing.

I do see the `Add New User with MySQL Auth Plugin` in the dropdown of the request samples. Are you expecting to see it somewhere else?

specification/resources/databases/databases_add_user.yml
```diff
@@ -38,6 +38,11 @@ requestBody:
             mysql_settings:
               auth_plugin: mysql_native_password
 
+        Add New Read Only User (MongoDB Only):
+          value:
+            name: my-readonly
+            readonly: true
```
danaelhe, line 44: I see the `Add New User with MySQL Auth Plugin` and `Add New User` are also documented in the example responsebody [here](https://github.com/digitalocean/openapi/blob/main/specification/resources/databases/responses/user.yml#L28)

Might be worth adding the `Add New Read Only User (MongoDB Only)` example responsebody as well.

danaelhe (commented):
Looks good! Have one small suggestion.

PR #808, opened 2023-06-12
Update Container Registry Endpoint Paths

danaelhe:
We had a Pydo user [open a github issue](https://github.com/digitalocean/pydo/issues/173) regarding the [list registry repository tags endpoint](https://docs.digitalocean.com/reference/api/api-reference/#operation/registry_list_repositoryTags). 

Turns out the client wasn't working because we were documenting the wrong endpoint path for the container registry endpoints, resulting in a "Not Found" 404 error.


andrewsomething (approved):
:man_facepalming: 

LGTM!

PR #809, opened 2023-06-14
Update databases_update_firewall_rules.yml

dbrian57:
Resolves PDOCS-1952. Added a missing bracket to the DBaaS firewall cURL example.


andrewsomething (approved):
:+1:  Thanks Dan

PR #811, opened 2023-06-29
Add Pydo Examples (about half of endpoints)

danaelhe:
This PR includes Pydo examples for the API Documentation. This first PR is about half of the endpoints. Will create another PR for the second half.

PR #812, opened 2023-06-30
Add Pydo Examples part 2

danaelhe:
This PR includes Pydo examples for the API Documentation. This second PR is the second half of the endpoints. First half was complete in #811


bentranter (approved):
Had a quick look at each example, LGTM!!

PR #814, opened 2023-07-23
Kubernetes: improve description and semantics on labels

timoreimann:
Make it clear that labels are not automatically removed from the corresponding Node objects if it is deleted in the API.

### DIFF
--- a/specification/resources/kubernetes/models/node_pool.yml
+++ b/specification/resources/kubernetes/models/node_pool.yml
@@ -66,8 +66,10 @@ kubernetes_node_pool_base:
       type: object
       nullable: true
       example: null
-      description: An object containing a set of Kubernetes labels. The keys and
-        are values are both user-defined.
+      description: An array of labels to apply to all nodes in a pool. Labels
+        will automatically be applied to all existing nodes and any subsequent
+        nodes added to the pool. Note that when a label is removed, it is not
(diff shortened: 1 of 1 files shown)


specification/resources/kubernetes/models/node_pool.yml
```diff
@@ -66,16 +66,18 @@ kubernetes_node_pool_base:
       type: object
       nullable: true
       example: null
-      description: An object containing a set of Kubernetes labels. The keys and
-        are values are both user-defined.
+      description: An array of labels to apply to all nodes in a pool. Labels
```
andrewsomething, line 69: > array of labels

Can we use a different term here? Think it's a bit confusing to call it an array when that is an OpenAPI data type, and this field is not actually that type.

timoreimann: Sounds fair. I'm not too familiar with OpenAPI data types, however -- would it be okay to call it "an object of key/value mappings" or something like that?

andrewsomething: :+1:  Sounds good

timoreimann: Updated -- please take another look when you get a chance @andrewsomething.

PR #815, opened 2023-08-01
Linting: Increase Incorrect OperationId Naming Severity to Error

danaelhe:
Increasing the spectral rule `operationid-must-follow-new-naming-conventions` from a warning to an error. This introduces a breaking change to Pydo as the `uptime` endpoints were not respecting the naming convention. We have two options we can take:

1) Accept the breaking changes and cut a new release of Pydo. Pydo hasn't had a 1.0 release, so could be fine.

2) Override the uptime checks paths to ignore the operationId naming convention as such:

```
  overrides:
  - files: ['**#/paths/~1v2~1uptime~1checks']
    rules:
      operationid-must-follow-new-naming-conventions: "off
```

Though slightly more disruptive, I prefer option 1. We are at an early stage in Pydo where a breaking change wouldn't be extremely impactful. Plus, I think the changes to the `uptime` functions feel better and are more consistent with the rest of the paths.

### DIFF
--- a/spectral/functions/validateOpIDNaming.js
+++ b/spectral/functions/validateOpIDNaming.js
@@ -9,9 +9,9 @@ const DELETE = ["delete", "destroy", "remove", "purge", "untag", "unassign"];
 const GET = ["get", "list"];
 const PATCH = ["patch"];
 const POST = ["create", "post", "add", "tag", "install", "reset", "upgrade",
-  "recycle", "run", "retry", "validate", "assign", "unassign", "cancel",
+  "recycle", "run", "retry", "validate", "assign", "unassign", "cancel", "list",
   "destroy", "delete", "update", "attach", "revert", "commit"];
-const PUT = ["update"];
+const PUT = ["update", "promote"];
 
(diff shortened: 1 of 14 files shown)


spectral/functions/validateOpIDNaming.js
```diff
@@ -9,9 +9,9 @@ const DELETE = ["delete", "destroy", "remove", "purge", "untag", "unassign"];
 const GET = ["get", "list"];
 const PATCH = ["patch"];
 const POST = ["create", "post", "add", "tag", "install", "reset", "upgrade",
-  "recycle", "run", "retry", "validate", "assign", "unassign", "cancel",
+  "recycle", "run", "retry", "validate", "assign", "unassign", "cancel", "list",
```
danaelhe, line 12: "List" was added because we have an endpoint that is best described as "listing" resources but it is indeed a POST request:
https://docs.digitalocean.com/reference/api/api-reference/#operation/apps_list_metrics_bandwidth_daily

andrewsomething (approved):
:+1:  LGTM!

PR #816, opened 2023-08-18
Add More Rate Limit Remaining Context

danaelhe:
Documenting that `Ratelimit-Remaining` reaching zero can also indicate that the 250 requests per minute limit was met, even if the 5,000 requests per hour limit was not. That way, users know they can hit a `Ratelimit-Remaining: 0` even if they haven't used up all 5,000 requests per hour.

Totally open to keeping this undocumented as well if this is common practice.


andrewsomething (approved):
:+1:  Good clarification

PR #828, opened 2023-10-07
fix: make period a required field for uptime alert

mikesmithgh:
Hi, I am working https://github.com/digitalocean/doctl/issues/1419 as part of #hacktoberfest and discovered `period` is not defined as a required field in the spec.

If I attempt to post a request without `period` then I receive
```json
{
    "id": "Internal Server Error",
    "message": "Server Error"
}
```

The field is also required on the UI.
uptime-alert-screenshot

### DIFF
--- a/specification/resources/uptime/create_alert.yml
+++ b/specification/resources/uptime/create_alert.yml
@@ -34,6 +34,7 @@ requestBody:
           - name
           - type
           - notifications
+          - period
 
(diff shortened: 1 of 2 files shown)


danaelhe:
GREAT catch! Will merge on green 🚀  thank you!

danaelhe (approved):
🚀

PR #833, opened 2023-10-15
APPS-7325 - deprecate app component routing/cors config

gregmankes:
users should use ingress rules instead

https://do-internal.atlassian.net/browse/APPS-7325


andrewsomething (approved):
:+1:  LGTM!

PR #834, opened 2023-10-16
Update poetry version used in Python client validation

andrewsomething:
Updates the poetry version used in Python client validation to match the one used by pydo itself.

See: https://github.com/digitalocean/pydo/pull/237

Resolves errors like:

```
  The lock file is not compatible with the current version of Poetry.
  Upgrade Poetry to be able to read the lock file or, alternatively, regenerate the lock file with the `poetry lock` command.
```

https://github.com/digitalocean/openapi/actions/runs/6526929286/job/17721187808#step:7:13


danaelhe (approved):
🚢

PR #839, opened 2023-10-26
Database User ACLS - make 'id' optional

dweinshenker:
`id` is not required to be provided when creating Kafka user acls. This property is computed and provided back in the response.


andrewsomething (approved):
:+1:  LGTM!

PR #845, opened 2023-11-29
Removes outdated database slugs

dbrian57:
This PR replaces outdated database slugs with recommendations on how to pull the latest list of slugs. Because the list is always changing, it is not practical to keep this list here until we figure out some kind of automated mechanism.


andrewsomething (approved):
:+1:

PR #850, opened 2024-01-22
Include DBAAS standby Nodes

dwilsondo:
The DO public API now includes standby node hostnames (standby_connection and standby_private_connection) in responses for clusters and pools. . This PR adds these fields to the OpenAPI spec.

Follows https://github.com/digitalocean/godo/pull/663


danaelhe (approved):
👍

PR #856, opened 2024-02-27
BILL-9714: Update docs for added invoice_id field

leodalbert:
Since we already expose the invoice_id as part of the BillingHistory endpoint as well as on the face of the invoice PDF, we should also include it here for consistency. 

See https://do-internal.atlassian.net/browse/BILL-9714


andrewsomething (approved):
:+1:  Thanks Leo!

PR #860, opened 2024-03-19
update error response for non-snapshots

icosahedral:
This doc update relates to a change in behavior in our public API related to a bug fix made back in January - details can be found in https://github.com/digitalocean/product-docs/pull/4150.  While working to migrate this endpoint to a new service, we found that image type unrelated to snapshots (for example backups and custom images) could be interacted with through these endpoints.  All these operations were properly scope to the user's account - the bug sole focused on the image type related to the record.  This was corrected, and we return a 400 in this scenario, with the following response:

```
{"id":"bad_request","message":"the resource is not a snapshot"}
```

### DIFF
--- a/specification/resources/snapshots/responses/not_a_snapshot.yml
+++ b/specification/resources/snapshots/responses/not_a_snapshot.yml
@@ -1,0 +1,18 @@
+description: Bad Request
+
+headers:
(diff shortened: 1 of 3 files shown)


andrewsomething (approved):
:+1:  LGTM!

PR #865, opened 2024-04-02
Updates Droplet name field with character limits

dbrian57:
Resolves some customer feedback we received about how the Droplet name field doesn't specify any limits. This PR updates the Droplet name field with the allowed number of characters and alpha-numeric pattern.

### DIFF
--- a/specification/resources/droplets/models/droplet_multi_create.yml
+++ b/specification/resources/droplets/models/droplet_multi_create.yml
@@ -7,6 +7,8 @@ allOf:
         type: array
         items:
           type: string
+          maxLength: 255
+          pattern: ^[a-z][a-z0-9-]{0,30}[a-z0-9]$
         example:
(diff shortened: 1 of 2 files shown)


specification/resources/droplets/models/droplet_multi_create.yml
```diff
@@ -7,6 +7,8 @@ allOf:
         type: array
         items:
           type: string
+          maxLength: 255
+          pattern: ^[a-z][a-z0-9-]{0,30}[a-z0-9]$
```
andrewsomething, line 11: These regexps seem overly restrictive. For example, the linter is failing as the examples do not match the pattern. `.` should be accepted as it is part of a valid hostname. Capital letters also seem to be accepted.

andrewsomething (approved):
Thanks Dan!

PR #867, opened 2024-04-04
Pin the Redoc version used in PR previews.

andrewsomething:
The live API docs are bundled at version 2.0.0-rc.50:

https://github.com/digitalocean/product-docs/blob/64ac3f9dce1bcc3749d9310c1623bdb5313951ff/assets/js/redoc.js#L4

Let's use the same one for our PR previews.


danaelhe (approved):
👍

PR #869, opened 2024-04-10
dbaas: metrics_endpoints is readOnly

andrewsomething:
Reviewing https://github.com/digitalocean/openapi/pull/868, I noticed that `metrics_endpoints` shows up as a create time option when it is not actually user configurable. It should be marked `readOnly`.


danaelhe (approved):
👍

PR #872, opened 2024-04-18
APPS-8386: Deprecate tier upgrade/downgrade fields from instance size

greeshmapill:
- Deprecates tier_downgrade_to and tier_upgrade_to from app instance size spec.
- Deprecates app_tier_downgrade_cost from propose app response.


andrewsomething (approved):
:+1:  LGTM, thanks!

PR #875, opened 2024-05-13
Add OpenAPI spec for dedicated egress IPs

ZachEddy:
This updates the OpenAPI spec for App Platform to include dedicated egress IPs.

Screenshot 2024-05-13 at 10 04 42 AM
Screenshot 2024-05-13 at 10 08 35 AM

### DIFF
--- a/specification/resources/apps/apps_create.yml
+++ b/specification/resources/apps/apps_create.yml
@@ -34,6 +34,8 @@ requestBody:
               instance_size_slug: apps-s-1vcpu-0.5gb
               routes:
                 - path: /api
+          egress:
+            type: DEDICATED_IP
   required: true
 
 responses:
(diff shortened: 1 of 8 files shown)


danaelhe:
Looks good! Looks like there's some linting errors that need to be addressed, you can view locally using `make lint`

Quick question- it appears the API is different from the UI. In the UI it appears you "add" dedicated egress IPs and then two IPs automatically get assigned and enabled. In the API, it appears the user has to input the specific IPs to be used? Is that correct?

ZachEddy:
@danaelhe Fixed those linting errors. Thanks for the tip about running `make lint` locally.

> In the API, it appears the user has to input the specific IPs to be used? Is that correct?

Not quite. The public API and UI behavior is the same. The user has to set `egress.type` to `DEDICATED_IP` in their app spec to enable the feature via the UI. The same is true of the public API. Those `192.168.1.x` IP addresses are included in the HTTP response part of the OpenAPI spec, not the request payload.

Having said that, your comment made me realize my PR was incomplete. I didn't include any OpenAPI spec changes that document how users can enable the feature. I added a change in 4b2fc82 that shows how users can create or update an app with Dedicated Egress enabled.

danaelhe (approved):
Thanks for the clarification! Looks great 🚀

PR #876, opened 2024-05-14
APPS-8732 - Add OpenSearch Log Forwarding destination.

ElanHasson:
This PR adds support for the OpenSearch Log Forwarding destination.


danaelhe (approved):
lgtm 👍

PR #882, opened 2024-06-04
APPS-8398: Mark ListTiers and GetTier as deprecated

greeshmapill:
After the Apps Relaunch in May, the new pricing structure is not tied to app tiers anymore. Hence, going forward, we would want to retire the use of tiers. 
(Note: The APIs will continue to work for a while until customer comms is done. We are just marking them as deprecated to discourage their usage.)


danaelhe (approved):
👍

PR #890, opened 2024-06-17
Add Log Forwarding Support for DBAAS

dweinshenker:
**What**

- Adds spec for log forwarding operations from DBAAS clusters
- Includes examples of elasticsearch, opensearch, and rsyslog


danaelhe (approved):
🚀

PR #898, opened 2024-07-25
Updates Database List Options response

dbrian57:
Resolves PDOCS-2597. Updates the response for the List Database Options endpoint with the latest engines, engine versions, and node sizes.


andrewsomething (approved):
:+1:

PR #900, opened 2024-07-29
Add OpenSearch Advanced Config for DBAAS

dweinshenker:
**What**

- Adds OpenSearch as an engine in main DBAAS description
- Adds OpenSearch advanced config to both PATCH/GET endpoints
- Refactors yml files to avoid naming overlap by scoping them, see https://github.com/Redocly/redocly-cli/issues/661


andrewsomething (approved):
:+1:  Looks great.

PR #906, opened 2024-09-05
apps: log_destinations is an array

andrewsomething:
`log_destinations` is an array. This is a "bug" with the spec, not a change to the API. For example, see godo:

https://github.com/digitalocean/godo/blob/9eb6e771f72ccc97ecfa8df04c433358d486588b/apps.gen.go#L508


danaelhe (approved):
Good catch 👍

PR #908, opened 2024-09-06
monitoring: add additional detail about metrics labels.

andrewsomething:
Adds a bit more to the `metric` description to make it more clear that the labels will vary depending on the metric being queried.


loosla (approved):
:+1:

PR #909, opened 2024-09-08
DBaaS | Add API endpoint for applying cluster patches

vsharma6855:


specification/resources/databases/databases_update_installUpdate.yml
```diff
@@ -0,0 +1,47 @@
+operationId: databases_update_installUpdate
```
andrewsomething, line 1: The operation ID is used by client generation tools to produce the method name. For example pydo, our Python client, will end up having a method with the redundant name `databases.update_install_update`. This might be more natural:

```suggestion
operationId: databases_install_update
```

Our linter enforces using consistent operation IDs across our API. As this would be the first PUT using `install`, you'll also need to add it here so that it passes:

https://github.com/digitalocean/openapi/blob/6fddf799ff4318258535091eed0eb7cd50bc6e6c/spectral/functions/validateOpIDNaming.js#L14

andrewsomething (changes requested):
Looks great, but I did have a suggestion in-line re: operationId

specification/resources/databases/databases_update_installUpdate.yml
```diff
@@ -0,0 +1,47 @@
...
+
+summary: Start Database Maintenance
+
+description: >-
+  To start the installation of updates now for a
+  database cluster, send a PUT request to
```
danaelhe, line 7:
```suggestion
  To start the installation of updates for a
  database cluster, send a PUT request to
```
I think removing the "now" might be easier to read?

specification/resources/databases/databases_update_installUpdate.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  A successful request will receive a 204 No Content status code with no body
+  in response.
+
+tags:
+  - Databases
+
+parameters:
+  - $ref: 'parameters.yml#/database_cluster_uuid'
+
```
loosla, line 18: I believe a single line in between should be enough.

loosla:
Looks good to me

specification/resources/databases/databases_update_installUpdate.yml
andrewsomething: I've pushed an additional commit to allow this to pass the linter check: https://github.com/digitalocean/openapi/pull/909/commits/e41ef07e319a3b74505140c43a88935fcb961c63

andrewsomething (approved):

PR #911, opened 2024-09-13
Document API changes supporting new Droplet backup policies

andrewsomething:


andrewsomething:
Will need a re-base on main once https://github.com/digitalocean/openapi/pull/912 lands.

specification/resources/droplets/models/droplet_actions.yml
```diff
@@ -25,6 +25,42 @@ droplet_action:
...
       description: The type of action to initiate for the Droplet.
 
+droplet_action_enable_backups:
+  allOf:
+  - $ref: '#/droplet_action'
+  - type: object
+    properties:
+      backup_policy:
+        allOf:
+        - $ref: 'droplet_backup_policy.yml'
```
loosla, line 35: :+1: Looks great.
Just curious if there is a way to clarify how the default backup policy looks like?

loosla (approved):
:+1:

specification/resources/droplets/models/droplet_actions.yml
andrewsomething: That's a good question! We should fast follow with some more details. @house-lee is the default weekly or daily?

PR #912, opened 2024-09-13
workflows: use actions/upload-artifact@v4

andrewsomething:
Our PR workflow has begun failing due to the deprecation of `actions/upload-artifact@v2`

> Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

https://github.com/digitalocean/openapi/actions/runs/10855754222/job/30129028044?pr=911

https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

This also removes a duplicate upload in same workflow due to a breaking change in v4. It was a duplicate and safe to remove.

> Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes

### DIFF
--- a/.github/workflows/spec.main.yml
+++ b/.github/workflows/spec.main.yml
@@ -33,15 +33,18 @@ jobs:
       - name: Bundle
         run: make bundle
 
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
           name: openapi-bundled
           path: tests/openapi-bundled.yaml
+
       - name: Set outputs
         id: vars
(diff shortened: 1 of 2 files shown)


loosla (approved):
Thank you :+1:

PR #914, opened 2024-09-20
PDOCS-2719: Clarify lbaas/tls_connections_per_second_utilization_percent alert requirements

beardicus:
Adds a note to the `v1/insights/lbaas/tls_connections_per_second_utilization_percent` alert type in two different tables clarifying that this alert only fires if the lbaas has an HTTPS forwarding rule.


andrewsomething (approved):
Thanks @beardicus!

PR #915, opened 2024-09-26
PDOCS-2781: update token scopes for firewall update operations

beardicus:
Updates six YAMLs to use the correct `firewall:update` scope.


andrewsomething (approved):
Thanks!

PR #917, opened 2024-09-27
Certificate token scope

dbrian57:
Resolves PDOCS-2782. Corrects the token scopes for certificate GET and certificate LIST endpoints.


andrewsomething (approved):
:man_facepalming:  Good catch! Thanks Dan!

PR #919, opened 2024-10-01
CDN rate limit clarification

dbrian57:
Resolves PDOCS-2746. Updates CDN doc strings with rate limit clarifications.


andrewsomething (approved):
:+1:  LGTM

andrewsomething:
This is covered up in the general rate limit section in the introduction, but documenting it closer to the specific endpoint makes sense.

PR #921, opened 2024-10-03
Fix outdated docs links

hazel-nut:
Noticed a few outdated/redirecting docs links, which I've replaced with a current link.


andrewsomething (approved):
:pray:  Thanks Hazel!

PR #924, opened 2024-10-09
VPC Peering Updates

beardicus:
Adds VPC peering info:

- List all peerings: GET `/v2/vpc_peerings`
- Create a new peering: POST  `/v2/vpc_peerings`

- Get a peering: GET `/v2/vpc_peerings/{vpc_peering_id}`
- Update a peering: PATCH `/v2/vpc_peerings/{vpc_peering_id}`
- Delete a peering: DELETE `/v2/vpc_peerings/{vpc_peering_id}`

- List a VPCs peerings: GET `/v2/vpcs/{vpc_id}/peerings`
- Add a peering to a specific VPC: POST `/v2/vpcs/{vpc_id}/peerings`

TODO:

- [x] I need somebody to check the auth token scopes for each of these
- [x] A thorough accuracy check from the engineering team would be good
- [x] A lookover from the API team would be a bonus

PR #925, opened 2024-10-10
vpcs: update operationId for vpcs_update_peerings

andrewsomething:
This is a follow on to https://github.com/digitalocean/openapi/pull/924. Reviewing https://github.com/digitalocean/pydo/pull/361, I noticed that it creates both  `VpcPeeringsOperations` and `VpcsPeeringsOperations`. This endpoint is under `/v2/vpcs` rather than `/v2/vpcs_peerings`.


loosla (approved):
:+1:

PR #926, opened 2024-10-10
Add `network` field for internal load balancers

beardicus:
Adds one field, `network` to load balancers.


andrewsomething (approved):
:+1:  LGTM

PR #927, opened 2024-10-11
Update two parameters to writeable during DOKS create

beardicus:
Updates `cluster_subnet` and `service_subnet` to be writeable during create only.

### DIFF
--- a/specification/resources/kubernetes/models/cluster.yml
+++ b/specification/resources/kubernetes/models/cluster.yml
@@ -32,14 +32,13 @@ properties:
   cluster_subnet:
     type: string
     format: cidr
-    readOnly: true
     example: 10.244.0.0/16
-    description: The range of IP addresses in the overlay network of the
+    description: The range of IP addresses for the overlay network of the
       Kubernetes cluster in CIDR notation.
 
   service_subnet:
     type: string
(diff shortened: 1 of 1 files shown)


specification/resources/kubernetes/models/cluster.yml
```diff
@@ -32,14 +32,13 @@ properties:
   cluster_subnet:
     type: string
     format: cidr
-    readOnly: true
     example: 10.244.0.0/16
```
gottwald, line 35: Please change these examples here and below because these are the two ranges they can specifically not choose since they're blocked for the traditional virtual networks:

Let's use something easy like:
`10.200.0.0/16` for pods and
`10.201.0.0/16` for services

gottwald: We could also keep it consistent with what you chose on the pdocs side:
```
"cluster_subnet": "192.168.0.0/20"
"service_subnet": "192.168.16.0/24"
```
Whatever you prefer.

beardicus: updated these to the ones i used in pdocs... thanks for the review!

andrewsomething (approved):

PR #931, opened 2024-10-15
actions: in-line region_slug property.

andrewsomething:
This in-lines the `region_slug` property in the actions response model. We had been overriding the `nullable` attribute using `allOf`. Some tools are not treating this as expected, inlcuding Redocly not displaying the example correctly: 

image

While the duplication isn't ideal, this is the simplest solution to unblock the pdocs team's work on pydo documentation.

CC: @hazel-nut


loosla (approved):
:rocket:

PR #937, opened 2024-10-27
fix: add accepted query params in docs for invoicesGetByUuid

lauhon:


andrewsomething (approved):
:+1:  Good catch! Thanks for the PR.

PR #939, opened 2024-10-29
Make sure to display old deprecated instance sizes

lauhon:
Hi,

I'm not entirely sure if this is valid. But I noticed that the old instance slugs still work, but are not part of the schema.

My idea is to display them, but as deprecated. This would be great for the project Im working on, because we still use those old slugs and would benefit from them being present in the API spec.


lauhon:
Hey @loosla, thanks for helping me out by approving the actions and merging main.

Could you check out the [failing pipeline](https://github.com/digitalocean/openapi/actions/runs/11578657143/job/32232992012?pr=939#step:3:29)?
It seems to have nothing to do with the file I changed, so I'm not sure how to fix it.

lauhon:
Thanks a lot @andrewsomething!

I applied your suggestions :)

PR #942, opened 2024-11-06
GET /v2/droplets/<id>/destroy_with_associated_resources{/status} require droplet:delete

andrewsomething:
This fixes the documentation for both `GET /v2/droplets/<id>/destroy_with_associated_resources` and `GET /v2/droplets/<id>/destroy_with_associated_resources/status`. Despite being GETs, they require the`droplet:delete` scope not `droplet:read`.


loosla (approved):
:+1:

PR #944, opened 2024-11-13
MultiRegistry Beta API

nayanjd-do:


sankalp-r (approved):

specification/resources/registry/examples/curl/all_registries_get.yml
```diff
@@ -0,0 +1,6 @@
+lang: cURL
+source: |-
+  curl -X GET \
+    -H "Content-Type: application/json" \
+    -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" \
+    "https://api.digitalocean.com/v2/registry"
```
andrewsomething, line 6:
```suggestion
    "https://api.digitalocean.com/v2/registries"
```

specification/resources/registry/multiregistry_create.yml
```diff
@@ -0,0 +1,51 @@
+operationId: multiregistry_create
```
andrewsomething, line 1:
```suggestion
operationId: registries_create
```

The operation ID is used by generated clients to name methods. So that we are consistent with godo and other resources, we probably want to use `registries` here in place of `multiregistry`.

specification/resources/registry/multiregistry_create.yml
```diff
@@ -0,0 +1,51 @@
...
+
+summary: Create Container Registry By Name (Beta)
```
andrewsomething, line 3:
```suggestion
summary: [Beta] Create Container Registry By Name
```

Let's put the beta tag up front for clarity.

specification/resources/registry/multiregistry_delete.yml
```diff
@@ -0,0 +1,39 @@
+operationId: multiregistry_delete
+
+summary: Delete Container Registry By Name (Beta)
```
andrewsomething, line 3:
```suggestion
operationId: registries_delete

summary: [Beta] Delete Container Registry By Name
```

specification/resources/registry/multiregistry_get.yml
```diff
@@ -0,0 +1,39 @@
+operationId: multiregistry_get
+
+summary: Get Container Registry By Name (Beta)
```
andrewsomething, line 3:
```suggestion
operationId: registries_get

summary: [Beta] Get a Container Registry By Name
```

specification/resources/registry/multiregistry_get_all.yml
```diff
@@ -0,0 +1,33 @@
+operationId: multiregistry_get_all
+
+summary: Get All Container Registries (Beta)
```
andrewsomething, line 3:
```suggestion
operationId: registries_list

summary: [Beta] List All Container Registries
```

specification/resources/registry/multiregistry_get_dockerCredentials.yml
```diff
@@ -0,0 +1,61 @@
+operationId: multiregistry_get_dockerCredentials
+
+summary: Get Docker Credentials By Registry Name (Beta)
```
andrewsomething, line 3:
```suggestion
operationId: registries_get_dockerCredentials

summary: [Beta] Get Docker Credentials By Registry Name
```

specification/resources/registry/examples/python/multiregistry_create.yml
```diff
@@ -0,0 +1,13 @@
...
+  from pydo import Client
+
+  client = Client(token=os.environ.get("DIGITALOCEAN_TOKEN"))
+
+  req = {
+    "name": "example",
+    "region": "fra1"
+  }
+
+  resp = client.registry.create(body=req)
```
andrewsomething, line 13:
```suggestion
  resp = client.registries.create(body=req)
```

As mentioned, these are generated from the operation ID. So at the moment, that would actually be `client.multiregistry.create`, but I think we really want `client.registries.create`.

specification/resources/registry/examples/curl/multiregistry_get.yml
```diff
@@ -0,0 +1,6 @@
+lang: cURL
+source: |-
+  curl -X GET \
+    -H "Content-Type: application/json" \
+    -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" \
+    "https://api.digitalocean.com/v2/registries"
```
andrewsomething, line 6:
```suggestion
    "https://api.digitalocean.com/v2/registries/example"
```

specification/resources/registry/models/multiregistry.yml
```diff
@@ -0,0 +1,39 @@
+type: object
```
andrewsomething, line 1: This is nearly identical to the existing registry model. It's just missing the `subscription` piece. I wonder if we could do this in a way that doesn't require duplication?

Maybe something like rename this to `models/registry_base.yml` and update the `models/registry.yml` to be:

```yml
type: object

allOf:
  - $ref: 'registry_base.yml'
  - type: object
    properties:
      subscription:
        allOf:
          - readOnly: true
          - $ref: 'subscription.yml'
```

andrewsomething (changes requested):
Overall this looks great. A few suggestions and questions inline.

specification/resources/registry/multiregistry_create.yml
nayanjd-do: Noted. I had to put it inside quotes or else the operation was not visible in the rendered spec.

specification/resources/registry/multiregistry_delete.yml
nayanjd-do: Done

specification/resources/registry/multiregistry_get.yml
nayanjd-do: Done

specification/resources/registry/multiregistry_get_all.yml
nayanjd-do: Done.

specification/resources/registry/multiregistry_get_dockerCredentials.yml
nayanjd-do: Done

specification/resources/registry/models/multiregistry.yml
nayanjd-do: Awesome suggestion @andrewsomething.  I have created `models/registry_base.yml` and updated `models/multiregistry.yml` and `models/registry.yml` to import from it. Lemme know if it looks good now.

specification/resources/registry/examples/python/multiregistry_create.yml
nayanjd-do: Got you @andrewsomething. Just for my information, is the python client generated automatically?

loosla: :wave: Python client is generated automatically as soon as `godo` changes merged to godo `main` branch.

nayanjd-do: Hey @andrewsomething @loosla 👋, we are back at this work.

> As mentioned, these are generated from the operation ID. So at the moment, that would actually be client.multiregistry.create, but I think we really want client.registries.create.

Just to confirm: if we want the python client to generate like `client.registries.xxx`, it is sufficient to change the `operationId`. I believe I should also change the filename to maintain consistency.

specification/resources/registry/registries_create.yml
```diff
@@ -0,0 +1,51 @@
+operationId: registries_create
+
+summary: "[Beta] Create Container Registry"
```
danaelhe, line 3: We're moving away from terms like "Beta" in favor of official terms from our product lifecycle:
https://docs.digitalocean.com/platform/product-lifecycle/

Please replace this with the term that is appropriate for where this feature is in the product lifecyle.

specification/resources/registry/examples/python/multiregistry_create.yml
kishlay-singh-DO: @nayanjd-do yes that is correct changing the `operationId` should be sufficient and yes changing the filename would be to a good choice too

danaelhe (dismissed):

specification/resources/registry/multiregistry_create.yml
nayanjd-do: Hey @andrewsomething, closing this as it has been updated.

specification/resources/registry/registries_create.yml
```diff
@@ -0,0 +1,51 @@
+operationId: registries_create
+
+summary: "[GA] Create Container Registry"
```
andrewsomething, line 3:
```suggestion
summary: "Create Container Registry"
```

If this API is generally available, it does not need to have a `[GA]` tag. That should be the default state. We only need to label endpoints that are not generally available so that customers understand that they may not have access.

rak16: I agree  - we should remove the `[GA]` tag. But I do think there should be some high-level indication to the user that there are two diff sets of API's in the list - the new (`/registries`) and the old (`/registry`). 

Is there a way we can tag the `/registry` endpoints without marking them as `deprecated`?

andrewsomething: > Is there a way we can tag the `/registry` endpoints without marking them as `deprecated`?

Short of marking them `deprecated`, the only way is updating the descriptions.

I'd suggest at least updating the top-level Container Registry description to include some information about the differences between the `/v2/registry` and `/v2/registries` endpoints:

https://github.com/digitalocean/openapi/blob/fe9078d917f5b311949397025701e9eab3e0901e/specification/DigitalOcean-public.v2.yaml#L132-L141

specification/DigitalOcean-public.v2.yaml
```diff
@@ -140,6 +140,14 @@ tags:
       You can only create one registry per DigitalOcean account, but you can use
       that registry to create as many repositories as you wish.
 
+  - name: "[Public Preview] Container Registry"
```
andrewsomething: One thing to be aware of here is that putting ` [Public Preview]` in the tag name will result in links like: `http://127.0.0.1:8080/#tag/Public-Preview-Container-Registry/operation/registries_list`

That means the links will break when they move into GA.

It might be better to just use

```
  - name: "Container Registries"
```

and add a sentence to the description stating that it is in Public Preview.

andrewsomething (approved):
Made a suggestion inline, but it is not blocking.

specification/DigitalOcean-public.v2.yaml
nayanjd-do: @andrewsomething Agreed. I have changed this. We would have to return the url  `http://127.0.0.1:8080/#tag/Container-Registries` from our backend which we would do next 👍

andrewsomething (approved):

PR #948, opened 2024-11-19
databases: set cluster_uuid in firewall_rule to readOnly.

andrewsomething:
The `cluster_uuid` attribute should be `readOnly`. While no error is returned, it is ignored in the request. On create, the new database cluster is used. On update, the UUID from the path is used.


loosla (approved):
Thank you :+1:

PR #953, opened 2024-12-05
Update apps_update.yml

dbrian57:
Updates App update cURL example to be complete.


specification/resources/apps/examples/curl/apps_update.yml
```diff
@@ -1,6 +1,5 @@
-lang: cURL
-source: |-
-  curl -X PUT \
-    -H "Content-Type: application/json" \
-    -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" \
-    "https://api.digitalocean.com/v2/apps/{id}"
+curl -X PUT \
```
andrewsomething, line 1:
```suggestion
lang: cURL
source: |-
  curl -X PUT \
```

This is a yaml file. The example needs to be under the `source` key.

andrewsomething (changes requested):
Good catch!

Let a comment inline that should resolve the CI failure. Beyond that, I wonder if we could use a more minimal example? This is pretty hard to read.

specification/resources/apps/examples/curl/apps_update.yml
dbrian57: ah, got it. Thanks!

dbrian57:
@andrewsomething sure thing! I work something up.

dbrian57:
@andrewsomething How do you feel about this? Should I uncompress it? I compressed it for easier copying, but maybe that's not the move here

andrewsomething (approved):
:+1:  Works for me. Thanks Dan!

PR #955, opened 2024-12-12
APPS-9365: Add bitbucket source to app spec

greeshmapill:
As per title.


loosla (approved):
Looks good to me :rocket:

PR #964, opened 2025-01-09
Fix required permissions for multiple endpoints

andrewsomething:
This fixes the required permissions for a number of endpoints. These are not due to any change. They were mis-documented.


mikeberezowski (approved):
Looks good to me 👍

PR #965, opened 2025-01-16
linter: update message and support additional verbs

andrewsomething:
This adds support for some additional verbs in the linter check on operation IDs.

It also updates a misleading error message. It's not really looking at the "end," it's looking at the first piece after the namespace:

https://github.com/digitalocean/openapi/blob/dc4c36f5568121f2502da2a9a7f32b4d63ec41ba/spectral/functions/validateOpIDNaming.js#L23

It was probably written when everything was simple like `droplets_list`.  For more complex names, we want consistency with existing ones like `databases_list_options`:

https://github.com/digitalocean/openapi/blob/dc4c36f5568121f2502da2a9a7f32b4d63ec41ba/specification/resources/databases/databases_list_options.yml#L1

### DIFF
--- a/spectral/functions/validateOpIDNaming.js
+++ b/spectral/functions/validateOpIDNaming.js
@@ -5,13 +5,13 @@
  *
  */
 
-const DELETE = ["delete", "destroy", "remove", "purge", "untag", "unassign"];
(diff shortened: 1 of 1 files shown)


loosla (approved):
:+1:

PR #968, opened 2025-01-20
Add field to customize the offline page during app maintenance

blesswinsamuel:
The `offline_page_url` allows users to customize the offline page when an app is in maintenance mode or when an app is archived.


markusthoemmes (approved):

thearyanahmed (approved):

loosla (approved):
:+1:

PR #974, opened 2025-01-28
LBAAS-3600: add ipv6 and network_stack field to public api docs

jvasilevsky:
add ipv6 and network_stack field to public api docs


andrewsomething (approved):
:+1:  LGTM

PR #976, opened 2025-02-04
Update kubernetes_upgrade_cluster.yml - k8s:update -> k8s:create

okamidash:
Our documentation is incorrect here, the correct scope is `kubernetes:create` rather than `kubernetes:update`


d-honeybadger (approved):

andrewsomething (approved):
:+1:  Good catch! Thanks!

PR #978, opened 2025-02-06
note on k8s RBAC for custom DO roles

d-honeybadger:
There are some gotchas on how Custom Roles work inside k8s clusters.


loosla (approved):
:+1: Looks good to me

PR #985, opened 2025-02-20
getagent parity with internal api

paperspaceBen:
this makes the getagent /v2/gen-ai/agents/<uuid> return the same data as the internal api endpoint. adding data that was missing for a customer


danaelhe (approved):
lgtm 👍

PR #986, opened 2025-03-04
Remove auth 1click list from docs

alexeigutium:
1-clicks list is a fully public endpoint and token isn't needed. This PR fixes documentation: token info is removed from list 1clicks information


danaelhe (approved):
good catch 👍

PR #989, opened 2025-03-11
APPS-10258: Update auth token scopes for apps APIs

greeshmapill:
Update token permissions on create app deployment and app exec endpoints to improve access control.

Note that, both the previous and the new permission for these endpoints are currently allowed. We choose to just go with the new permissions in the doc to make sure users creating a new token use the ones that will work in the future.
The old permissions will be retired after customer comms and on completion of notice period following that.


loosla (approved):
:+1:

PR #991, opened 2025-03-18
APPS-10258: Update auth token scope for get exec API

greeshmapill:
Missed to update scope for app exec per deployment API in https://github.com/digitalocean/openapi/pull/989


loosla (approved):
:rocket:  Thank you

PR #958, opened 2025-01-02
Add Spaces Key API

lee-aaron:
OBJECT-6022

This adds information about the Spaces Key API to the OpenAPI spec.

I can add a `[Beta]` tag somewhere if needed as users would need to be on the flipper to access this!

PR #993, opened 2025-03-20
Add Spaces Key API Updated

amahajan-do:
OBJECT-6022

This adds information about the Spaces Key API to the OpenAPI spec.

This PR supersedes #958


andrewsomething (approved):
:+1:  LGTM, thanks for the updates!

For posterity, here is the diff against https://github.com/digitalocean/openapi/pull/958

PR #998, opened 2025-03-26
Add Partner Network Connect

beardicus:
Adds Partner Network Connect. Releases April 1.

### DIFF
--- a/specification/resources/partner_network_connect/partner_attachment_create.yml
+++ b/specification/resources/partner_network_connect/partner_attachment_create.yml
@@ -1,6 +1,11 @@
-operationId: create_partner_attachment
+operationId: attachments_create
 
 summary: Create a new partner attachment
+
+description: |
+  To create a new partner attachment, send a `POST` request to 
+  `/v2/partner_network_connect/attachments` with a JSON object containing the
+  required configuration details.
 
 tags:
   - Partner Network Connect
--- a/specification/resources/partner_network_connect/partner_attachment_create.yml
+++ b/specification/resources/partner_network_connect/partner_attachment_create.yml
@@ -25,6 +25,9 @@ responses:
 
   '404':
     $ref: '../../shared/responses/not_found.yml'
+  
+  '422':
+    $ref: '../../shared/responses/unprocessable_entity.yml'
 
   '429':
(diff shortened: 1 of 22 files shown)


specification/resources/partner_network_connect/partner_attachment_create.yml
```diff
@@ -0,0 +1,40 @@
...
+responses:
+  '202':
+    $ref: 'responses/single_partner_attachment.yml'
+
+  '401':
+    $ref: '../../shared/responses/unauthorized.yml'
+
+  '404':
+    $ref: '../../shared/responses/not_found.yml'
+
```
loosla, line 28: Could you add 422 response code for create?

I was able to get it with this in response:
`"id":"unprocessable_entity","message":"invalid NaaS Provided"
`

beardicus: added the shared `unprocessable_entity` blurb.

specification/resources/partner_network_connect/partner_attachment_delete.yml
```diff
@@ -0,0 +1,36 @@
...
+  `/v2/partner_network_connect/attachments/{pa_id}`.
+
+tags:
+  - Partner Network Connect
+
+parameters:
+  - $ref: 'parameters.yml#/pa_id'
+
+responses:
+  '202':
```
loosla, line 16: Could you double check please if we have a response body on delete please?
Usually the successful Delete returns 204 without any body in response.

loosla (dismissed):
:rocket: Please address the comment about Delete :pray:

specification/resources/partner_network_connect/partner_attachment_delete.yml
beardicus: I just checked and it does return a response body where `"state": "DELETING"`.

loosla: Thanks a lot for check it.
Could you update the response then? :pray: 

That's how it looks for me (doesn't have "state", but includes multiple other fields which I suppose are not there)
Screenshot from 2025-03-26 20-05-16

loosla (approved):
Thank you :rocket:

PR #999, opened 2025-03-31
Update kubernetes API spec: cluster status messages

kperath:
tested locally with `make  preview`

### DIFF
--- a/specification/resources/kubernetes/responses/status_messages.yml
+++ b/specification/resources/kubernetes/responses/status_messages.yml
@@ -1,0 +1,16 @@
+description: |
+  The response is a JSON object which contains status messages for a Kubernetes cluster. Each message object contains a timestamp and an indication of what issue the cluster is experiencing at a given time.
+
+
+headers:
+  ratelimit-limit:
+    $ref: '../../../shared/headers.yml#/ratelimit-limit'
+  ratelimit-remaining:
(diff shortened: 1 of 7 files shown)


specification/resources/kubernetes/responses/status_messages.yml
```diff
@@ -0,0 +1,16 @@
...
+    $ref: '../../../shared/headers.yml#/ratelimit-limit'
+  ratelimit-remaining:
+    $ref: '../../../shared/headers.yml#/ratelimit-remaining'
+  ratelimit-reset:
+    $ref: '../../../shared/headers.yml#/ratelimit-reset'
+
+content:
+  application/json:
+    schema:
+      $ref: '../models/status_messages.yml'
```
andrewsomething, line 16: This is an array or `messages` not a single message, right?

```suggestion
      properties:
        messages:
          type: array
          items:
            $ref: "../models/status_messages.yml"
```

andrewsomething (changes requested):
Overall looks great, but one question in-line.

specification/resources/kubernetes/responses/status_messages.yml
kperath: Ah yes ty!!!

andrewsomething (approved):
:+1:  LGTM

PR #1001, opened 2025-04-02
Update operation IDs for partner attachment endpoints

andrewsomething:
This updates the operation IDs for the partner attachment endpoints so that the generated pydo client follows the same naming scheme as we used in godo.


loosla (approved):
Thanks Andrew :pray:

PR #1006, opened 2025-04-25
APPS-11069: Add spec for app health API

greeshmapill:
Exposes new http API to fetch health of app components.


loosla (approved):
Excellent job :rocket:  Thank you

PR #1007, opened 2025-04-28
Get Partner Attachment via ID 

anup-deka:
Added curl, go and python examples for get partner attachment via id

### DIFF
--- a/specification/resources/partner_network_connect/models/partner_attachment.yml
+++ b/specification/resources/partner_network_connect/models/partner_attachment.yml
@@ -93,6 +93,13 @@ partner_attachment:
         - ee1ad867-7026-44a5-bfcc-2587b52e1291
       description: An array of associated partner attachment UUIDs.
       x-order: 12
+    redundancy_zone:
+      type: string
+      format: string
+      readOnly: true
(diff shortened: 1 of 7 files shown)


specification/resources/partner_network_connect/models/partner_attachment.yml
```diff
@@ -93,6 +93,13 @@ partner_attachment:
         - ee1ad867-7026-44a5-bfcc-2587b52e1291
       description: An array of associated partner attachment UUIDs.
       x-order: 12
+    redundancy_zone:
```
loosla, line 96: Could you confirm with the network team if it's ok to add `redundancy_zone` to the model and it looks correct (as Andrew advised on a call)? :pray: 

The rest looks great! Good job.

If you want, you can just remove changes for `redundancy_zone` and create a separate PR for it when you get a response from the network team. It's up to you. Ping me again if you want to go this way, I'll approve your PR then.

anup-deka: Yeah, Sure I will add changes related to `redundancy_zone ` in another PR

PR #1018, opened 2025-05-05
Adding examples for regenerate service key for partner attachment

kishlay-singh-DO:
This PR updates the examples for regenerate service key for partner attachments.

https://github.com/user-attachments/assets/d48f9e62-803c-4c60-af77-f8f76f90211c

### DIFF
--- a/specification/resources/partner_network_connect/examples/curl/partner_attachment_service_key_create.yml
+++ b/specification/resources/partner_network_connect/examples/curl/partner_attachment_service_key_create.yml
@@ -1,0 +1,6 @@
+lang: cURL
+source: |-
+  curl -X POST \
+    -H "Content-Type: application/json" \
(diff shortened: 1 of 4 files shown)


loosla (approved):
:rocket: Excellent work! Thank you

This is slightly out of scope, but could you please confirm with the network team whether the `RegenerateServiceKey` response body is indeed empty? If that's the case, could you remove the response body in a separate PR?

Really appreciate it 🙏

kishlay-singh-DO:
> 🚀 Excellent work! Thank you
> 
> This is slightly out of scope, but could you please confirm with the network team whether the `RegenerateServiceKey` response body is indeed empty? If that's the case, could you remove the response body in a separate PR?
> 
> Really appreciate it 🙏

Sure will check with the team!

PR #1021, opened 2025-05-06
DBAAS: Update spec with rbac changes for mongo

dyunkim:
Response schema for `GET /users/$username` (similar output for `GET /users/` to list all users)
image

Request schema for `POST /users`
image

With new example for `POST /users`
image


dweinshenker (approved):

loosla (approved):
Thanks a lot :rocket:

PR #1023, opened 2025-05-08
Add documentation for instance_name for GetExec()

thearyanahmed:
We introduced a new request field `instance_name`, this is optional. This field allows us to get an exec url to a specific instance, beneficial for apps with multiple instances.

This PR adds documentations for that.


harshmaru7 (approved):
LGTM !

PR #1032, opened 2025-05-27
new fields

paperspaceBen:
some new fields recently added

### DIFF
--- a/specification/resources/gen-ai/definitions.yml
+++ b/specification/resources/gen-ai/definitions.yml
@@ -1,3 +1,26 @@
+apiAWSDataSource:
+  description: AWS S3 Data Source
+  properties:
+    bucket_name:
+      description: Spaces bucket name
+      example: '"example name"'
+      type: string
+    item_path:
+      example: '"example string"'
+      type: string
+    key_id:
+      description: The AWS Key ID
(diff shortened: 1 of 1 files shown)


specification/resources/gen-ai/definitions.yml
```diff
@@ -1,3 +1,26 @@
+apiAWSDataSource:
+  description: AWS S3 Data Source
+  properties:
+    bucket_name:
+      description: Spaces bucket name
+      example: '"example name"'
+      type: string
+    item_path:
+      example: '"example string"'
```
kishlay-singh-DO, line 9: Can we please add a description to the `item_path` property as well and the other properties where its missing.

specification/resources/gen-ai/definitions.yml
```diff
@@ -1,3 +1,26 @@
...
+      type: string
+    key_id:
+      description: The AWS Key ID
+      example: '"123e4567-e89b-12d3-a456-426614174000"'
+      type: string
+    region:
+      description: Region of bucket
+      example: '"example string"'
+      type: string
```
danaelhe, line 18: I'm assuming this is AWS region? Maybe can update example and description so it's more clear?

PR #1034, opened 2025-05-30
SSH key clarification

dbrian57:
Clarifies that SSH keys must already exist on your team in order to be added to a Droplet.


andrewsomething (approved):
Thanks!

PR #1037, opened 2025-06-04
Add app platform edge controls

thearyanahmed:
Relevant release note https://docs.digitalocean.com/products/app-platform/#19-may-2025


specification/resources/apps/models/app_spec.yml
```diff
@@ -26,6 +26,27 @@ properties:
     - syd
     example: nyc
 
+  disable_edge_cache:
+    description: |-
+      If set to `true`, the app will not be cached at the edge (CDN). This is useful if you want to bring your own CDN configuration,
```
ElanHasson: when we say "not available for static sites", do you mean Static Site components or Apps with static sites ?

I suspect the latter because they share the same ingress?

specification/resources/apps/models/app_spec.yml
```diff
@@ -26,6 +26,27 @@ properties:
...
+  disable_email_obfuscation:
+    description: |-
+      If set to `true`, email addresses in the app will not be obfuscated. This is
+      useful for apps that require email addresses to be visible (in the HTML markup).
+    type: boolean
+    default: false
+
+  enhanced_threat_control_enabled:
+    description: |-
+      If set to `true`, suspicious requests will go through additional security checks to help mitigate layer 7 DDoS attacks.
```
ElanHasson, line 48: I think we should add a link to pdocs because this doesn't explain what it really does

specification/resources/apps/models/app_spec.yml
```diff
@@ -26,6 +26,27 @@ properties:
...
+    description: |-
+      If set to `true`, the app will not be cached at the edge (CDN). This is useful if you want to bring your own CDN configuration,
+      whether through external CDN provider or in app caching, and for
+      apps that require real-time data or have dynamic content, using SSE (server-sent events) with GET methods or hosting your MCP (Model Context Protocol) Server. Note: This feature is not available for static sites.
+    type: boolean
+    default: false
+
+  disable_email_obfuscation:
+    description: |-
+      If set to `true`, email addresses in the app will not be obfuscated. This is
```
ElanHasson, line 41: we should also link to pdocs here to further clarify this.

specification/resources/apps/models/app_spec.yml
```diff
@@ -26,6 +26,27 @@ properties:
     - syd
     example: nyc
 
+  disable_edge_cache:
+    description: |-
+      If set to `true`, the app will not be cached at the edge (CDN). This is useful if you want to bring your own CDN configuration,
+      whether through external CDN provider or in app caching, and for
+      apps that require real-time data or have dynamic content, using SSE (server-sent events) with GET methods or hosting your MCP (Model Context Protocol) Server. Note: This feature is not available for static sites.
```
ElanHasson: we should also link to pdocs here to further clarify this.

1. I'm not sure what "bring your own CDN configuration" or "external cdn provider" means in this context
2. with "app caching", do you mean caching output at the app level? If so, that would be a different concern that wouldn't matter if edge cache was enabled or disabled here.  The two caches play a different role in optimization.
3. I'm also unsure of why we mentioned MCP here also. It's oddly specific and doesn't seem directly relevant to the spec function, aside from MCP being able to use a websocket or sse transport.

specification/resources/apps/models/app_spec.yml
thearyanahmed: I'm not sure about the difference of Static Site Components or Apps with static sites. If you are referring "Apps with static sites" to those apps, that has one or multiple `service` components, and they can choose to serve static content, then it's the first one. 

This feature is not available for apps with `Static Site` components.

thearyanahmed: Updated the `Note` section to `This feature is not available for static site components.`

specification/resources/apps/models/app_spec.yml
thearyanahmed: We'll have to re-iterate after PDOCS adds this section.

specification/resources/apps/models/app_spec.yml
thearyanahmed: We'll have to re-iterate after PDOCS adds these section.

specification/resources/apps/models/app_spec.yml
thearyanahmed: For the first point, I wanted to express that users who also use say Cloudflare or Akamai (or any other CDN provider) to front their domain, as of now, either those requests are going through double CDN or a they can't use their own configuration and just let the request go through their CDN to our Origin. 

This feature enables our users to allow disable AP CDN cache. So their own CDN to AP Origin. 

Kindly suggest if you have better wording. 

And for point 2 and 3, I've updated the whole comment and linked PDOCS.

specification/resources/apps/models/app_spec.yml
ElanHasson: Thanks for clarifying the language here. My position on docs is that we should be as clear as possible about the way we say things so that it doesn't set incorrect expectations of users and generate support cases when it doesn't work as they understood.

specification/resources/apps/models/app_spec.yml
```diff
@@ -26,6 +26,27 @@ properties:
     - syd
     example: nyc
 
+  disable_edge_cache:
+    description: |-
+      If set to `true`, the app will not be cached at the edge (CDN). This is useful if you want to bring your own CDN configuration,
+      whether through external CDN provider or caching and serving static content with in your app, and for
+      apps that require real-time data or have dynamic content, using SSE (server-sent events) with GET methods or hosting your MCP (Model Context Protocol) Server that uses SSE. Note: This feature is not available for static site components. Refer to the [Disable CDN Cache](https://docs.digitalocean.com/products/app-platform/how-to/cache-content/#disable-cdn-cache) for more information.
```
ElanHasson:
```suggestion
      If set to `true`, the app will **not** be cached at the edge (CDN). Enable this option if you want to manage CDN configuration yourself—whether by using an external CDN provider or by handling static content and caching within your app. This setting is also recommended for apps that require real-time data or serve dynamic content, such as those using Server-Sent Events (SSE) over GET, or hosting an MCP (Model Context Protocol) Server that utilizes SSE.  
      **Note:** This feature is not available for static site components.  
      For more information, see [Disable CDN Cache](https://docs.digitalocean.com/products/app-platform/how-to/cache-content/#disable-cdn-cache).

```

ElanHasson: I think this may flow a bit better. thoughts?

thearyanahmed: thanks. I prefer your version and committed.

PR #1038, opened 2025-06-05
VPC-4475, VPC-4437: update two Partner Network Connect endpoints

beardicus:
Updates partner attachment "create service key" docs to reflect its empty JSON response and async nature.
Updates partner attachment "create" docs with new `redundancy_zone` parameter.

Addresses VPC-4475 and VPC-4437.

### DIFF
--- a/specification/resources/partner_network_connect/responses/empty_json_object.yml
+++ b/specification/resources/partner_network_connect/responses/empty_json_object.yml
@@ -12,7 +12,7 @@ headers:
 content:
   application/json:
     schema:
-      - type: object
-        properties: {}
+      type: object
+      properties: {}
     example: {}
 


andrewsomething (approved):
:pray:  Thanks!

specification/resources/partner_network_connect/models/partner_attachment.yml
```diff
@@ -160,6 +160,11 @@ partner_attachment_writable:
           type: string
           example: 0xsNnb1pwQlowdoMySEfWwk4I
           description: BGP Auth Key
+    redundancy_zone:
+      type: string
+      description: Optional redundancy zone for the partner attachment.
+      enum: [ UNSPECIFIED, MEGAPORT_BLUE, MEGAPORT_RED ]
```
lfundaro-do, line 166: the UNSPECIFIED doesn't have to be added I think. They can just leave this field as empty and the server will get UNSPECIFIED by default. Saying this so we don't create confusion on the user having to set that.

PR #1039, opened 2025-06-06
remove UNSPECIFIED from partner attachment enum

beardicus:
Removes one incorrect option from an enum for a partner attachment, as per @lfundaro-do.


lfundaro-do (approved):

specification/resources/partner_network_connect/models/partner_attachment.yml
```diff
@@ -163,7 +163,7 @@ partner_attachment_writable:
     redundancy_zone:
       type: string
       description: Optional redundancy zone for the partner attachment.
```
kishlay-singh-DO, line 167: QQ:- If we are removing `UNSPECIFIED ` from the enum shouldn't this now not be optional?

SSharma-10 (approved):

PR #1041, opened 2025-06-13
APPS-11421: Add spec for app templates list API

fumblehool:
Exposes new http API to fetch app templates list.


specification/resources/apps/models/app_template.yml
```diff
@@ -0,0 +1,45 @@
+type: object
+properties:
+  name:
+    type: string
```
harshmaru7, line 4: can we add a description for each of the properties ?

PR #1053, opened 2025-07-03
New Spectral Function Added: Ensuring Schema property has type defined

Mahendra-Chittupolu:
[APCLI-3150](https://do-internal.atlassian.net/browse/APICLI-3150)

- some schema defined in openApi spec missing **type** fields which causing **Kiota**(typescript sdk generator) to send out nested value fields response.
- to resolve this issue implemented **ensureSchemaHasType.js** acts as a linter and logs the fields if they do not have a  **type** field.

Example:

```
openapi/specification/resources/monitoring/monitoring_update_destination.yml
 36:14  error  schema-properties-must-have-type  Schema property is missing 'type' field. Path: paths./v2/monitoring/sinks/destinations/{destination_uuid}.post.requestBody.content.application/json.schema.properties.type  requestBody.content.application/json.schema

```

### DIFF
--- a/spectral/functions/ensureSchemaHasType.js
+++ b/spectral/functions/ensureSchemaHasType.js
@@ -1,0 +1,56 @@
+/**
+ * Copyright 2020 Box, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


anup-deka (dismissed):
LGTM

spectral/functions/ensureSchemaHasType.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright 2020 Box, Inc. All rights reserved.
```
andrewsomething: If this is borrowed from open source code from Box, can we link to the source. This makes it easier for us to update it in the future, e.g. 

https://github.com/digitalocean/openapi/blob/28e9ed174e7a011719ce5fe219ab17244906b929/spectral/functions/ensureAllArraysHaveItemTypes.js#L20-L21

PR #1054, opened 2025-07-04
docr: Fix registry create payload model for multi-registry

rak16:
The registry create model for mult-registry was missing `subscsription_tier_slug` - added that.


harshmaru7 (approved):
LGTM

gane5hvarma (approved):

PR #1058, opened 2025-07-10
Add GenAI evaluation endpoints and update docs

tgillam-do:
Updates the GenAI documentation:

- Adds evaluation endpoints
- Updates GenAI Platform -> GradientAI where it wasn't already

Screenshot 2025-07-10 at 8 24 08 AM


specification/resources/gen-ai/examples/curl/genai_delete_knowledge_base_data_source.yml
```diff
@@ -1,7 +1,7 @@
 lang: cURL
 source: |-
-  curl -X DELETE \
+  curl -X POST \
```
SSharma-10: why did you change this from `DELETE` to `POST`?

tgillam-do: Good catch. This is what we have in the internal docs but definitely is wrong. Reverted.

SSharma-10:
LGTM!

SSharma-10 (approved):

PR #1059, opened 2025-07-14
APPS-11763: VPC integration

ShivaniKumar1:
This PR adds vpc fields to AppSpec to allow VPC integration for App Platform.


ZachEddy (commented):
We should add the `vpc` field to the app spec request payload for [creates](https://docs.digitalocean.com/reference/api/digitalocean/#tag/Apps/operation/apps_create) and [updates](https://docs.digitalocean.com/reference/api/digitalocean/#tag/Apps/operation/apps_update). Other than that, LGTM!

specification/resources/apps/apps_create.yml
```diff
@@ -36,6 +36,10 @@ requestBody:
                 - path: /api
           egress:
             type: DEDICATED_IP
+          vpc:
+            id: c22d8f48-4bc4-49f5-8ca0-58e7164427ac
+            egress_ips:
+              - ip: 10.0.0.1
```
thearyanahmed: this is a read only property, won't be part of the request body but response.

specification/resources/apps/examples/python/apps_update.yml
```diff
@@ -225,6 +225,14 @@ source: |-
                 "version": "12",
             }
         ],
+        “vpc”: [
+            {
+                “id”: “c22d8f48-4bc4-49f5-8ca0-58e7164427ac”
+                “egress_ips”: {
```
thearyanahmed: Same comment as above

specification/resources/apps/models/apps_vpc.yml
```diff
@@ -0,0 +1,12 @@
+type: object
+readOnly: true
+properties:
+  id:
+    title: The ID of the VPC. 
+    type: string
+    example: c22d8f48-4bc4-49f5-8ca0-58e7164427ac
+  egress_ips:
```
thearyanahmed, line 8: from the PR, it's  (often) kinda difficult to see where this is being used. If it's in the request body, I guess then you'll have to update this file to exclude `egress_ips`.

specification/resources/apps/models/apps_vpc_egress_ip.yml
```diff
@@ -0,0 +1,5 @@
+type: object
+properties:
+  ip:
+    type: string
+    example: 10.0.0.1
```
thearyanahmed, line 5: This is not being reused, we can a dedicated file for it.

thearyanahmed (changes requested):

specification/resources/apps/models/apps_vpc.yml
ShivaniKumar1: This is for the response.

specification/resources/apps/examples/python/apps_update.yml
```diff
@@ -225,6 +225,11 @@ source: |-
                 "version": "12",
             }
         ],
+        “vpc”: [
```
thearyanahmed: Do we support multiple VPCs per app? As far as I know, **no**. And in that case, it should be a single object. 
 
```py
"vpc": {
    "id": ... 
}
```

thearyanahmed (approved):
nice work 🙌🏽

PR #1064, opened 2025-07-23
add docs for new doks cluster options

m3co-code:
Those options are already available in prod. So could be merged right away.


SSharma-10 (approved):
LGTM!

PR #1065, opened 2025-07-23
fix incorrect property names in API desc and add new fields to cluster_read

m3co-code:
Mistake spotted by @SSharma-10 (ty again!) in https://github.com/digitalocean/godo/pull/877


specification/resources/kubernetes/models/cluster_read.yml
```diff
@@ -165,6 +165,12 @@ properties:
   routing_agent:
     $ref: "routing_agent.yml"
 
+  amd_gpu_device_plugin:
```
m3co-code, line 168: Overlooked that before on the read part.

PR #1069, opened 2025-07-28
Add GET/UPDATE kafka schema/subject config

bhardwajRahul:
#### Schema Registry Connection 
image

#### Get Kafka Schema Registry Config
image

#### Update Kafka Schema Registry Config
image

#### Get Kafka Schema Registry Subject Config
image

#### Update Schema Registry Subject Config 
image

### DIFF
--- a/specification/resources/databases/databases_update_kafka_schema_subject_config.yml
+++ b/specification/resources/databases/databases_update_kafka_schema_subject_config.yml
@@ -1,10 +1,10 @@
-operationId: databases_update_kafka_schema_config
+operationId: databases_update_kafka_schema_subject_config
 
-summary: Update Schema Registry Configuration for a kafka Cluster
+summary: Update Schema Registry Configuration for a Subject of kafka Cluster
 
 description: |
-  To update the Schema Registry configuration for a Kafka cluster, send a PUT request to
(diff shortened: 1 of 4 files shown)


specification/resources/databases/databases_update_kafka_schema_subject_config.yml
```diff
@@ -0,0 +1,60 @@
...
+
+  "500":
+    $ref: "../../shared/responses/server_error.yml"
+
+  default:
+    $ref: "../../shared/responses/unexpected_error.yml"
+
+security:
+  - bearer_auth:
+      - "database:write"
```
kishlay-singh-DO, line 60: LGTM. I just wanted to ask this:-
Will the customer see the `bearer_auth` and role of `database:write `and would clearly understand this? or should we explicitly add it in the description?

bhardwajRahul: I don't think we need to add it explicitly in the description as same is being used everywhere in the api doc.

bhardwajRahul: specification/resources/databases/databases_get_config.yml

PR #1077, opened 2025-08-13
Add nat gateways spec

asaha2:


SSharma-10:
Hey @asaha2 , there are multiple OpenAPI spec errors causing the test case to fail. Could you please look into and resolve them?

SSharma-10 (approved):
LGTM

PR #1080, opened 2025-08-20
Add byoip patch prefix, remove public preview tag from byoip 

imaskm:


specification/resources/byoip_prefixes/responses/byoip_prefix_update.yml
```diff
@@ -0,0 +1,20 @@
...
+  headers:
+    ratelimit-limit:
+      $ref: '../../../shared/headers.yml#/ratelimit-limit'
+    ratelimit-remaining:
+      $ref: '../../../shared/headers.yml#/ratelimit-remaining'
+    ratelimit-reset:
+      $ref: '../../../shared/headers.yml#/ratelimit-reset'
+
+  content:
+    application/json:
```
vnegi-digitalocean, line 13: If I'm correct, its implementation is still pending?

anup-deka (approved):
Looks good 👍 
Removed the public preview and added the update operation

PR #1091, opened 2025-09-17
CON-12531 Add DOKS multi-registries documentation

llDrLove:
Add the multi-registries DOKS documentation for the `POST` and `DELETE` api calls. For now, this is only exposed using the public api and not integrated into the go, python and ruby clients.


DO-rrao (commented):
Hey,
I have been able to add multiple registries successfully.
However, I am not able to hit the endpoint - Remove Container Registries from Kubernetes Clusters
Can you send a screenshot with a 200 response for this endpoint?
otherwise, LGTM!

SSharma-10 (approved):
LGTM

PR #1140, opened 2026-02-24
update region slug enum

a-talsma:
This PR updates the region slug enum.


specification/shared/attributes/region_slug.yml
```diff
@@ -16,6 +16,6 @@ enum:
   - sfo2
   - sfo3
   - sgp1
-  - tor1
   - syd1
+  - tor1
```
a-talsma, line 19: alpha sorting

anup-deka (approved):

specification/shared/attributes/region_slug.yml
anup-deka: Hmm, some checks are failing

PR #1145, opened 2026-03-12
CON-13682 - HA default for 1.36.0 and later

pyadagiri-do:
CON-13682:
Updates the `ha` field default to `true` for Kubernetes clusters on version 1.36.0 and later. Documents version-dependent default behavior in cluster, cluster_read, and cluster_update models.


sjana-do (approved):
LGTM

PR #1153, opened 2026-04-03
Extends coredns_autoscaler field in Kubernetes cluster

skumarc-do:
New model: specification/resources/kubernetes/models/coredns_autoscaler.yml
- Referenced from:
  - models/cluster.yml
  - models/cluster_read.yml
  - models/cluster_update.yml
- Examples updated in specification/resources/kubernetes/responses/examples.yml where other plugin blocks are shown.

Referenced PRs - https://github.com/digitalocean/godo/pull/986

### DIFF
--- a/specification/resources/kubernetes/models/cluster.yml
+++ b/specification/resources/kubernetes/models/cluster.yml
@@ -177,6 +177,9 @@ properties:
   rdma_shared_dev_plugin:
     $ref: "rdma_shared_dev_plugin.yml"
 
+  coredns_autoscaler:
+    $ref: "coredns_autoscaler.yml"
+
 required:
   - name
   - region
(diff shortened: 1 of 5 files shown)


RohithSangati:
For all create, update, and delete requests, will the coredns_autoscaler field always be present in the response from now onwards, regardless of whether its value is null or not?

RohithSangati (approved):

skumarc-do:
> For all create, update, and delete requests, will the coredns_autoscaler field always be present in the response from now onwards, regardless of whether its value is null or not?

Yes. The cluster body returned by Create/Update/Get/List always includes coredns_autoscaler (defaults to { "enabled": false }) because publicapi.ToCluster initializes it before serialization, same as the other plugin fields (routing_agent, amd_*, nvidia_*, rdma_*).

PR #1157, opened 2026-04-20
APPS-12973: Update spec to include request metrics for autoscaling and alerts

greeshmapill:
Also split autoscaling spec into separate yml for service and non-service components.


specification/resources/apps/models/app_job_spec.yml
```diff
@@ -4,6 +4,9 @@ allOf:
 
 - type: object
   properties:
+    autoscaling:
+      $ref: app_autoscaling_spec.yml
+
```
blesswinsamuel: I think autoscaling is not supported for jobs.

greeshmapill: You are right, updated it accordingly, thanks.

blesswinsamuel (approved):

harshmaru7 (approved):
LGTM !

PR #1158, opened 2026-04-21
add rerank configs in controlplane

snegi-do:
https://github.internal.digitalocean.com/digitalocean/cthulhu/pull/157530

### DIFF
--- a/specification/resources/gen-ai/definitions.yml
+++ b/specification/resources/gen-ai/definitions.yml
@@ -1602,6 +1602,8 @@ apiCreateKnowledgeBaseInputPublic:
       description: The datacenter region to deploy the knowledge base in.
       example: '"tor1"'
       type: string
+    reranking_config:
+      $ref: '#/apiRerankingConfiguration'
     tags:
       description: Tags to organize your knowledge base.
       example:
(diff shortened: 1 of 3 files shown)


specification/resources/gen-ai/definitions.yml
```diff
@@ -4585,6 +4589,18 @@ apiReleaseStatus:
   - RELEASE_STATUS_BUILDING
   example: RELEASE_STATUS_UNKNOWN
   type: string
+apiRerankingConfiguration:
+  description: Configuration for cross-encoder reranking during retrieval.
+  properties:
```
harshmaru7, line 4594: just to validate - there are no required fields?

specification/resources/gen-ai/definitions.yml
```diff
@@ -4585,6 +4589,18 @@ apiReleaseStatus:
...
+    enabled:
+      description: Whether reranking is enabled for retrieval
+      example: true
+      type: boolean
+    model:
+      description: Reranker model internal name
```
harshmaru7, line 4600: can we re-phrase this ? not sure what internal refers to here

specification/resources/gen-ai/definitions.yml
snegi-do: no

specification/resources/gen-ai/definitions.yml
snegi-do: we require the internal name of model. Users can view the internal names of all models in the model catalog UI page. Model catalog shows internal name of models. I've also given example of internal name (currently we only have one reranker model)

harshmaru7 (approved):

PR #1161, opened 2026-04-22
Add v1/messages and v1/embeddings

SSharma-10:
Added support for the /v1/messages and /v1/embeddings endpoint


harshmaru7 (approved):
**LGTM**

PR #1166, opened 2026-04-28
APPS-13774: add custom error page URL to app ingress

Gborgman05:
Added field to reflect new functionality in app specification.

### DIFF
--- a/specification/resources/apps/models/app_ingress_spec.yml
+++ b/specification/resources/apps/models/app_ingress_spec.yml
@@ -5,4 +5,11 @@ properties:
     type: array
     items:
       $ref: app_ingress_spec_rule.yml
+  custom_error_page_url:
+    type: string
+    description: |-
+      Optional HTTPS URL of a custom error page to display when the app is unreachable. The
+      page is shown in a full-viewport iframe. If omitted, the default platform error page
+      is used.
+    example: https://status.example.com/app-unavailable
 description: Specification for app ingress configurations.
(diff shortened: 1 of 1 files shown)


specification/resources/apps/models/app_ingress_spec.yml
```diff
@@ -5,4 +5,11 @@ properties:
     type: array
     items:
       $ref: app_ingress_spec_rule.yml
+  custom_error_page_url:
```
Bala-Nallamilli, line 8: Hey, quick clarification before I approve — the PR title says "error **redirects**" but the description on this field says the page is rendered in a full-viewport **iframe**. Those are pretty different behaviors:

- **Redirect (302)** → browser navigates away to the URL, user leaves our domain.
- **Iframe embed** → user stays on our error page, customer's URL is loaded inside a frame.

Which one is the platform actually doing? Want to make sure the field name, description, and PR title all line up. If it's truly an iframe, `custom_error_page_url` is fine. If it's a redirect, something like `error_redirect_url` might read more accurately.

Also two minor things while I'm here:

- Since the description says HTTPS-only, worth tightening the schema with `format: uri` (and maybe `pattern: '^https://'`) so SDK consumers know up front instead of finding out via a 4xx.
- If it's an iframe, a one-liner noting the target page must allow framing (no `X-Frame-Options: DENY` / restrictive CSP) would save support tickets later.

Nothing blocking — happy to approve once the redirect-vs-iframe bit is sorted.

Gborgman05: Hey, Happy to clarify:
This is not a redirect, this is an embedded iframe, the user stays on our error page and the customer's URL is loaded if defined. I will update the PR title to clarify. 

I will implement the two suggestions to improve here. 
Thanks for the feedback!

Bala-Nallamilli (approved):
LGTM

harshmaru7 (approved):

PR #1170, opened 2026-05-13
Update DI update curl to latest deployment spec

alekhyarao1218-wq:
Updated the Dedicated Inference update curl example to align with the latest deployment spec structure.

Testing

- Verified YAML formatting.

- Compared update example with latest create example for consistency.


guptado (approved):

RohithSangati (approved):
LGTM

PR #1175, opened 2026-05-20
add worker_subnet_uuid to kubernetes docs

d-honeybadger:
New field was just added to DOKS API.

### DIFF
--- a/specification/resources/kubernetes/examples/curl/kubernetes_create_cluster.yml
+++ b/specification/resources/kubernetes/examples/curl/kubernetes_create_cluster.yml
@@ -3,6 +3,5 @@ source: |-
   curl -X POST \
     -H "Content-Type: application/json" \
     -H "Authorization: Bearer $DIGITALOCEAN_TOKEN" \
-    -d '{"name": "prod-cluster-01","region": "nyc1","version": "1.14.1\
-    -do.4","tags": ["production","web-team"],"node_pools": [{"size": "s-1vcpu-2gb","count": 3,"name": "frontend-pool","tags": ["frontend"],"labels": {"service": "frontend", "priority": "high"}},{"size": "c-4","count": 2,"name": "backend-pool"}]}' \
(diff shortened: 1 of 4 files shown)


Bala-Nallamilli:
A few questions/suggestions before approving:

* Is worker_subnet_uuid immutable after cluster creation? I noticed cluster_update.yml isn't touched, which suggests this can't be changed post-create. If that's intentional, could we call it out in the description on the create side? Similar to how the ha field documents its behavioral constraints. Something like: "This value cannot be changed after the cluster is created."

* Should there be a scope requirement note? The sibling vpc_uuid field ends its description with Requires \vpc:read` scope.Since subnets are a VPC sub-resource, does setting/readingworker_subnet_uuidalso requirevpc:read(or some other scope)? Worth aligning with thevpc_uuid` precedent if so.

* Create-request example doesn't demonstrate the new field. The response examples in responses/examples.yml include worker_subnet_uuid, but the curl create example in examples/curl/kubernetes_create_cluster.yml doesn't show a caller actually setting it. Could we add it there so users see how to use the new field end-to-end?

* Minor: The description says "Requires vpc_uuid to be set" — is this constraint enforced server-side with a validation error? Just confirming, since the schema itself doesn't express the dependency (no dependentRequired/oneOf).

Otherwise LGTM — examples coverage is complete and the write/read description split follows the existing pattern nicely.

d-honeybadger:
> Is worker_subnet_uuid immutable after cluster creation? 

Yes, added a callout

> Should there be a scope requirement note? 

Yes, TY! Required vpc:read, same as vpc_uuid

> Create-request example doesn't demonstrate the new field.

Didn't realize that was separate from the response examples tbh. Added the field there.

> The description says "Requires vpc_uuid to be set" — is this constraint enforced server-side with a validation error? Just confirming, since the schema itself doesn't express the dependency (no dependentRequired/oneOf).

Yes, there is a validation in the API for this. I tried adding `dependentRequired`, but apparently it's only available in openapi 3.1 and we're on 3.0. And using "oneOf" to express this seems confusing, no? Not even sure how to express this with oneof tbh. If it's okay to just have a description callout, I'd keep it this way.

Bala-Nallamilli (approved):
* Immutability note added ✅
* vpc:read scope documented on both create and read sides ✅
* Curl create example now demonstrates the new field (and the broken line-continuation fix is a nice bonus) ✅
* Constraint wording softened ✅

Nice call on keeping worker_subnet_uuid in only some of the response examples — showing both the "custom subnet" and "default subnet" cases is clearer than having every example set it.

LGTM

PR #1184, opened 2026-06-08
ESC-23439 | Added Missing API Documentation Values

SSharma-10:


specification/resources/autoscale_pools/models/member.yml
```diff
@@ -17,8 +17,18 @@ properties:
     example: 2020-07-28T18:00:00Z
   health_status:
```
DO-rrao, line 18: responses/examples.yml (members_all, lines 69–83): the rendered example for the list-members response was not updated alongside the schema. Since unhealthy_reason is documented as always returned (null when healthy), the example should include it (e.g. unhealthy_reason: null) so the docs stay in sync with the schema.

specification/resources/autoscale_pools/models/member.yml
```diff
@@ -17,8 +17,18 @@ properties:
...
     type: string
```
DO-rrao, line 19: models/member.yml line 25: unhealthy_reason is not added to required. The description says it "Returns null when health_status is healthy" — i.e. the field is always present, just sometimes null. With nullable: true it should be listed in required to convey "always returned, may be null".

SSharma-10: Updated

specification/resources/autoscale_pools/models/member.yml
```diff
@@ -17,8 +17,18 @@ properties:
...
-    example: 'active'
```
DO-rrao, line 20: Hi shivani, 
I ran the following request for healthy autoscale pool -http://127.0.0.1:8080/#tag/Droplet-Autoscale-Pools/operation/autoscalepools_list_members . I did not see null field of unhealthy_reason.

Can you pls confirm this is expected behaviour?
For unhealthy autoscale pools, I am seeing number of droplets is becoming 0. 
Then I am not able to see the field again.

SSharma-10: Yes you are correct,  unhealthy_reason: null not present for healthy droplets and I have updated the changes. Please check

DO-rrao (approved):

PR #1192, opened 2026-06-22
Nfs access point apis

sgupta832:
## Summary

Adds NFS access point CRUD endpoints to the public OpenAPI spec, aligned with the cthulhu sharesvc/publicapi implementation.

- `POST /v2/nfs/shares/{share_id}/access_points` — create access point
- `GET /v2/nfs/shares/{share_id}/access_points` — list access points (optional status filter)
- `GET /v2/nfs/access_points/{access_point_id}` — get access point
- `DELETE /v2/nfs/access_points/{access_point_id}` — delete access point
- `access_points` array on share GET/LIST responses

Gen1 scope: Create, Get, List, Delete only. `UpdateAccessPointPolicy` is intentionally omitted (not yet supported in sharesvc).

Supersedes #1191 — includes original spec plus review fixes (validation docs, correct delete action status, shared action schema, error responses).

## Test plan

- [x] `make bundle` passes
- [x] `make lint` passes (0 errors)

### DIFF
--- a/specification/resources/nfs/models/access_point_action_response.yml
+++ b/specification/resources/nfs/models/access_point_action_response.yml
@@ -1,0 +1,53 @@
+---
+type: object
+description: Response returned after starting an access point mutation.
+properties:
+  access_point:
+    $ref: "access_point_response.yml"
+  action:
+    type: object
+    description: The action that was submitted for the access point mutation.
+    properties:
+      id:
+        type: string
+        description: The unique identifier of the action.
+        example: "1"
+      region_slug:
+        type: string
+        description: The DigitalOcean region slug where the resource is located.
+        example: "atl1"
+      resource_id:
+        type: string
+        format: uuid
+        description: The unique identifier of the resource on which the action is being performed.
+        example: "a1b2c3d4-e5f6-4a5b-9c8d-1e2f3a4b5c6d"
+      resource_type:
+        type: string
+        enum: ["network_file_share"]
+        description: The type of resource on which the action is being performed.
+        example: "network_file_share"
+      started_at:
+        type: string
+        format: date-time
+        description: The timestamp when the action was started.
+        example: "2025-10-14T11:55:31.615157397Z"
+      status:
+        type: string
+        enum: ["in-progress", "completed", "errored"]
+        description: The current status of the action.
+        example: "in-progress"
+      type:
+        type: string
+        description: The type of action being performed. Access point mutations use CREATE_ACCESS_POINT and DELETE_ACCESS_POINT.
+        example: "CREATE_ACCESS_POINT"
+    required:
+      - region_slug
+      - resource_id
+      - resource_type
+      - started_at
+      - status
+      - type
+required:
+  - access_point
+  - action
+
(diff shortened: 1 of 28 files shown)


v-amanjain-afk (approved):

specification/resources/nfs/models/access_point_action_response.yml
```diff
@@ -0,0 +1,11 @@
+---
```
DO-rrao, line 1: Please add x-codeSamples on all four new operations(atleast one would be good too).

specification/resources/nfs/responses/access_point_create.yml
```diff
@@ -0,0 +1,43 @@
...
+              anonuid: 65534
+              anongid: 65534
+              protocols: ["NFS4"]
+              squash_config: "ROOT_SQUASH"
+              identity_enforcement_enabled: true
+          action:
+            id: "1"
+            region_slug: "atl1"
+            resource_id: "a1b2c3d4-e5f6-4a5b-9c8d-1e2f3a4b5c6d"
+            resource_type: "network_file_share"
```
DO-rrao: Small thing that confused me here: resource_id is the access point UUID but resource_type is network_file_share, and there's no access-point value in the enum over in nfs_action.yml. Is that what the API actually returns? Just want to make sure we're not mislabeling the action — same thing shows up in the delete response too.

specification/resources/nfs/nfs_access_point_create.yml
```diff
@@ -0,0 +1,74 @@
...
+              protocols: ["NFS4"]
+              squash_config: "ROOT_SQUASH"
+              identity_enforcement_enabled: true
+
+responses:
+  '201':
+    $ref: 'responses/access_point_create.yml'
+
+  '400':
+    $ref: 'responses/bad_request.yml'
```
DO-rrao: The 400 here points at bad_request.yml, but that one's all about share sizing (size_gib >= 50Gib), which doesn't really fit access point creation. Could we give it its own 400 example (or at least a more generic message)? Otherwise the docs will tell people the wrong thing. Delete has the same issue on its 400.

specification/resources/nfs/nfs_access_point_create.yml
```diff
@@ -0,0 +1,74 @@
+operationId: nfs_create_access_point
+
+summary: Create an NFS access point
+
```
DO-rrao, line 4: can you pls share response for this endpoint as is?
will help me to review.

specification/resources/nfs/models/access_point_action_response.yml
sgupta832: Added cURL x-codeSamples on all four AP operations.

specification/resources/nfs/responses/access_point_create.yml
sgupta832: Fixed — resource_type is SHARE, status is IN_PROGRESS; examples and nfs_action.yml updated.

specification/resources/nfs/nfs_access_point_create.yml
sgupta832: Replaced share bad_request.yml with access_point_bad_request.yml on create/delete.

specification/resources/nfs/nfs_access_point_create.yml
sgupta832: Updated create (and list/get/delete) examples to match staging — pls see latest commit.

DO-rrao (approved):
lgtm

PR #1196, opened 2026-06-30
Add custom evaluation metrics OpenAPI endpoints

sshirolkar:
## Summary
- add OpenAPI operations for custom evaluation metrics create/update/delete under the GenAI section
- align request/response schemas and shared metric fields with the Cthulhu proto-generated contract
- add curl examples for create, update (path metric UUID), and delete (path metric UUID)

## Test plan
- [x] Run `npm run lint -- specification/DigitalOcean-public.v2.yaml`
- [x] Confirm no linter errors in modified files via IDE diagnostics
- [x] Verify branch diff is scoped to custom metric OpenAPI files only

Made with [Cursor](https://cursor.com)

### DIFF
--- a/specification/resources/gen-ai/definitions.yml
+++ b/specification/resources/gen-ai/definitions.yml
@@ -1275,9 +1275,9 @@ apiAnthropicAPIKeyInfo:
   type: object
 apiAssociatedModelEvaluationPreset:
   description: |-
-    AssociatedModelEvaluationPreset identifies a saved model evaluation preset
-    that references a metric. Returned alongside custom metrics so the dashboard
-    can warn that deleting the metric will also delete these presets.
+    A saved model evaluation preset that references a metric. Returned alongside
+    custom metrics so you can see which presets depend on the metric; deleting the
+    metric will also delete these presets.
   properties:
     eval_preset_uuid:
       description: Unique identifier of the saved model evaluation preset that references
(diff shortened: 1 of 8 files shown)


specification/resources/gen-ai/definitions.yml
```diff
@@ -2734,6 +2766,42 @@ apiCustomModelImportJob:
...
+    Configuration for a custom model-evaluation metric scored by an LLM judge.
+    Prompt and model response are always included in the judge context.
+  properties:
+    add_to_library:
+      description: |-
+        When true, persist in the account metric library for reuse (Figma "Always show this metric").
+        When false, use for the current evaluation flow only. Set on create/update via config.
+      example: true
+      type: boolean
+    created_at:
```
DO-rrao, line 2774: One thing worth a second look: this config schema is shared by the create input, the update input, and the response (`apiEvaluationMetric.custom_eval_config`). That means `created_at`, `updated_at`, and `deleted_at` show up as fields someone can send in a create/update request, which probably isn't what we want people to do. Could we mark these three as `readOnly: true`, or split out a separate input config so the timestamps only appear on the response side?

specification/resources/gen-ai/definitions.yml
```diff
@@ -2734,6 +2766,42 @@ apiCustomModelImportJob:
...
       type: string
   type: object
+apiCustomEvaluationMetricConfig:
+  description: |-
+    Configuration for a custom model-evaluation metric scored by an LLM judge.
+    Prompt and model response are always included in the judge context.
+  properties:
+    add_to_library:
+      description: |-
+        When true, persist in the account metric library for reuse (Figma "Always show this metric").
```
DO-rrao: Small thing, but the `(Figma "Always show this metric")` reference will end up in our public docs and won't mean anything to external developers. Mind dropping it? While we're here, the `deleted_at` ("must not appear in pickers") and `associated_presets` ("so the dashboard can warn…") descriptions lean a bit on internal/dashboard behavior too — might be worth rephrasing those in terms of what the API consumer actually sees.

specification/resources/gen-ai/definitions.yml
sshirolkar: marked as readonly

DO-rrao (approved):

PR #1202, opened 2026-07-15
DBAAS-8634: Document pg_allow_replication optional default and update path

shabtaisharon:
Clarify that PostgreSQL user settings.pg_allow_replication defaults to false when omitted, and document create/update/read paths in the Public API spec.


RohithSangati (approved):

RohithSangati:
Ignoring gitguardian failure as it's due to addition of example password.

PR #1204, opened 2026-07-22
teamgroups api list and create team openapi spec

bking-do:
adds openapi spec for new teamgroup/organization endpoints

### DIFF
--- a/specification/DigitalOcean-public.v2.yaml
+++ b/specification/DigitalOcean-public.v2.yaml
@@ -419,6 +419,14 @@ tags:
       NFS actions are tasks that can be executed on an NFS share. These can be
       things like resizing, snapshotting, etc.
 
+  - name: Organizations
+    description: |-
+      Organizations group multiple teams under a shared administrative boundary.
+      The Organizations API lets you create teams within an organization and
+      list the teams that belong to it.
+
+      These endpoints must be called in an organization context.
+
   - name: Partner Network Connect
     description: |-
       Partner Network Connect lets you establish high-bandwidth, low-latency
@@ -690,6 +698,7 @@ x-tagGroups:
       - Monitoring
       - NFS
       - NFS Actions
+      - Organizations
       - Partner Network Connect
       - Project Resources
       - Projects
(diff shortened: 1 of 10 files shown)


specification/DigitalOcean-public.v2.yaml
```diff
@@ -1994,6 +2003,14 @@ paths:
     delete:
       $ref: "resources/nfs/nfs_access_point_delete.yml"
 
+  /v2/organizations/team:
+    post:
+      $ref: "resources/organizations/organizations_create_team.yml"
+
+  /v2/organizations/teams:
```
DO-rrao, line 2010: The create path /v2/organizations/team is singular while the list path /v2/organizations/teams is plural. Can you confirm this matches the backend API?

specification/DigitalOcean-public.v2.yaml
```diff
@@ -1994,6 +2003,14 @@ paths:
...
+    get:
```
DO-rrao, line 2011: This list endpoint has no page/per_page params and the response doesn't use the standard allOf + shared/pages.yml#/pagination + shared/meta.yml pagination pattern. Is the collection intentionally unpaginated? If not, please add pagination for consistency with other list endpoints.

specification/resources/organizations/models/organization_team.yml
```diff
@@ -0,0 +1,106 @@
+organization_team:
+  type: object
+  description: A team that belongs to an organization.
+  properties:
+    id:
+      type: integer
+      format: uint64
```
DO-rrao, line 7: If status and team_invitation.role have a fixed set of backend values, adding an enum (like team_invitation_status) would make the docs clearer.

specification/DigitalOcean-public.v2.yaml
bking-do: yes, this matches the backend api

specification/DigitalOcean-public.v2.yaml
bking-do: these backends are not paginated at the moment, i will look into adding pagination quickly

specification/resources/organizations/models/organization_team.yml
bking-do: yep, good call, added

DO-rrao (approved):
LGTM

PR #1210, opened 2026-07-31
Hide Secrets Manager from public API docs until GA

MaxLivesInTheMatrix:
## Summary
- Temporarily comment out Secrets Manager paths (`/v2/security/secrets*`) in the public OpenAPI root so they are not published to docs.digitalocean.com
- Keep the Secrets Manager operation/model/example files under `resources/security/` so they can be restored by uncommenting
- Update the Security tag description to CSPM-only while Secrets Manager remains unpublished

Customers are discovering Secrets Manager via public docs and receiving 403s because the product is not generally available.

## Test plan
- [x] `make lint` passes locally
- [x] Confirm Redoc/docs preview no longer lists Secrets Manager endpoints under Security
- [x] Confirm CSPM Security endpoints still appear
- [ ] After merge, verify https://docs.digitalocean.com/reference/api/reference/security/ no longer shows List/Get/Create/Update/Delete Secret endpoints

Made with [Cursor](https://cursor.com)

### DIFF
--- a/specification/DigitalOcean-public.v2.yaml
+++ b/specification/DigitalOcean-public.v2.yaml
@@ -540,9 +540,10 @@ tags:
       An action object is returned. These objects hold the current status of the
       requested action.
 
+  # Secrets Manager omitted from this tag description until GA; paths commented out below.
   - name: Security
     description: |-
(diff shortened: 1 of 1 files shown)


anup-deka (approved):
LGTM

PR #1211, opened 2026-08-03
doks: document the isolated_workers field

varshavaradarajan:
The field is only available to be set on k8s cluster create. And has prerequisite of NAT GW in the same VPC as cluster for isolated workers to be supported.


anup-deka (approved):
LGTM

PR #1215, opened 2026-08-14
Add DRA and gpu partition mode for GPU

thearyanahmed:
This PR includes DRA plugin api spec and `gpu_partition_mode` specs for DOKS.


d-honeybadger (approved):
👍 from DOKS

RohithSangati (approved):

PR #1216, opened 2026-09-01
updated specs for byoip

rbhatia-code:


specification/resources/vpc_nat_gateways/models/vpc_nat_gateway_create.yml
```diff
@@ -71,6 +71,31 @@ properties:
...
+    properties:
+      public_gateways:
+        type: array
+        description: >-
+          An array containing a single public gateway object that sets the
+          gateway's public egress IP address.
+        items:
+          type: object
+          properties:
+            ip:
```
SSharma-10, line 94: GET/list currently document `egresses.public_gateways[].ipv4`, but create here only documents `ip`

If the API accepts `ip` on create (and maybe also `ipv4`) but returns `ipv4`, we should document that explicitly so create and get don’t look inconsistent to generated clients

specification/resources/load_balancers/models/load_balancer_create.yml
```diff
@@ -10,6 +10,10 @@ oneOf:
         region:
           $ref: '../../../shared/attributes/region_slug.yml'
 
+    - type: object
+      properties:
+        $ref: 'attributes.yml#/load_balancer_ip'
```
SSharma-10: This schema is also used by `loadBalancers_update`. Adding a writable `ip` here means the update API now exposes a field that the description says is create-only and immutable

Previously, `ip` was `readOnly` in `load_balancer_base.yml`, so clients would not treat it as a request field. This looks like a contract change for generated clients/docs

Can we keep `ip` writable only for create, either with a separate update schema or by keeping it `readOnly` for update?

SSharma-10 (approved):
LGTM

